PR #24, opened 2020-07-25
Add flake8, ignore migrations

tony:


alexmojaki:
It might make sense to ignore the chapters entirely, although there's no need to jump to that. Sometimes steps have intentionally wrong code in them.

Did this not raise any warnings to fix?

Issue #23, opened 2020-04-10
Next course chapters


alexmojaki:
Superseded by #92 

The biggest thing to do is write the actual course!

This should be an informal, collaborative process. The course doesn't need to be written in the order that it's going to be presented, we can change things as we go, and any amount of brainstorming is useful. Even sharing a little code snippet to demonstrate a concept or a neat exercise is helpful. [See below](#issuecomment-665624422) for an example.

If you want to contribute a solid bit of course with actual text, code, and exercises, open a new issue with a proposal just in markdown. Don't jump to implementing it in Python.

Before writing anything, it's a good idea to go through the course to know what's been covered and to get a feel for the style. You don't have to actually read everything or do all the exercises - see the [developer mode instructions](https://github.com/alexmojaki/futurecoder#controls) to skip steps. Reading all the hints on a few exercises would also help.

It can be helpful to read through existing textbooks and such to find inspiration for material. Make sure you have permission to copy their ideas (even if you rephrase it into different words) or that it's permissively licensed. If you're not sure, ask them. I don't want anyone to accuse us of plagiarism.

I'm not sure what exactly the next chapters should be at this stage, but here are some loosely related topics that I want to cover:

- ~How to use the birdseye debugger - this will be helpful for understanding `and` and `or` correctly.~ Done
- ~Boolean operators `and`, `or`, and `not` - these will be helpful for constructing more useful conditions for `if` and `while`.~ (currently being worked on in #27)
    - ~Must cover common misuse, e.g. `if x == 1 or 2` instead of `if x == 1 or x == 2`. Birdseye can help here.~
- `while`. Start with a simple counter (contrast with `for i in range`), then another example with a proper condition (any ideas for a compelling example?), then `while True`.
- Basics of types:
    - Different values have different types
    - These can be revealed with the `type()` function
    - Types include `str`, `int`, `float`, `bool`, and `list`.
    - Different types behave differently, e.g. `+` and `<` means different things for each, and each have different methods.
    - Different types are often 'incompatible', e.g. numbers and strings cannot be added, and `==` is always False.
    - How to convert between types, e.g. str to int and vice versa.
- ~`input()` - goes well with `while`. Explain that it always returns a string, and must convert to int when needed.~ Has a volunteer in #61 
- Nested lists

spamegg1:
Hey @alexmojaki 
I discovered this from your OSSU post. 
This is just the kind of thing I was looking for. 
[...]

PR #32, opened 2020-08-04
feature(naming_rule): present naming_rule of python in the excercise of c03_variables

BobAnkh:
## Description

Change the expression from `name` to `your_name` in page `Using Variables` of c03_variables.py

Ref: #23 

## Types of changes

- New feature (non-breaking change which adds functionality)

## Other information

I only change one page of this chapter because I think it is enough to present the rule

### DIFF
--- a/backend/main/chapters/c03_variables.py
+++ b/backend/main/chapters/c03_variables.py
@@ -66,49 +66,51 @@ Previously we made a variable called `word` with the value `'Hello'` with this c
 
     word = 'Hello'
 
-Now make a variable called `name` whose value is another string. The string can be anything...how about your name?
+Now make a variable called `your_name` whose value is another string. The string should be your name. Try it!
+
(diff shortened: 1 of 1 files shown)


backend/main/chapters/c03_variables.py
```diff
@@ -66,49 +66,51 @@ class name_assign(Step):
 
     word = 'Hello'
 
-Now make a variable called `name` whose value is another string. The string can be anything...how about your name?
+Now make a variable called `your_name` whose value is another string. The string should be your name. Try it!
+
+TIPS: The character `_` in `your_name` is called an *underscore*. Use it to separate words when you want a variable name containing multiple words. You can type it on most keyboards by pressing Shift and hyphen/dash/minus (`-`).
```
alexmojaki, line 71:
```suggestion
Now make a variable called `your_name` whose value is another string.

(The character `_` in `your_name` is called an *underscore*. Use it to separate words when you want a variable name containing multiple words. You can type it on most keyboards by pressing Shift and hyphen/dash/minus (`-`).)
```

alexmojaki:
Great, thanks! I'll merge this later today.

BobAnkh:
😄

PR #40, opened 2020-08-08
Encourage editor instead of shell to solve list insert exercise

alexmojaki:


BobAnkh:
Will we use `Good find! Let's do one more. Consider this program(run it from the editor)` instead of `Good find! Let's do one more. Consider this program` makes clearer?

alexmojaki:
I think that:

1. It's safe to assume that multiline programs are understood to be meant for the editor.
2. They don't need to actually run the example, they should know what it does and we tell them anyway.
3. It's fine if they complete the exercise in the shell, it should still work.

BobAnkh:
👍

Issue #28, opened 2020-08-01
Process dies trying to run Python command on Docker


opheron:
While runnning locally on Docker, when trying to run 1+2 on "Introducing The Shell", the in-browser command prompt returned:

`The process died.
Your code probably took too long.
Maybe you have an infinite loop?
`

This was on first install. Now, on second run, I can't seem to reproduce the problem.

alexmojaki:
Thanks for the report, I appreciate the effort. But since you can't reproduce it and neither can I, I think it makes sense to close this for now. Please reopen if you find more information.

alexmojaki:
@opheron I have implemented open chapters/pages as we discussed in #38. Please see the description and try out the test environment.

PR #49, opened 2020-08-12
Try to make system more robust to random failures by also recreating queues when restarting processes

alexmojaki:
Trying to deal with the sort of problem mentioned in #28 or #47. If someone could try fiddling with the deployment to test that'd be great - @BobAnkh?

### DIFF
--- a/backend/main/workers/master.py
+++ b/backend/main/workers/master.py
@@ -5,7 +5,7 @@ import queue
 from collections import defaultdict, deque
 from functools import lru_cache
 from multiprocessing import Process, Queue
-from threading import Thread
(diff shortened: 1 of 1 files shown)


BobAnkh:
I will try it when the deployment is ready.

BobAnkh:
LGTM. I haven't seen these issues. But only try the deployment with codes in editor and commands in shell. I haven't tested it locally using a docker.

My Environment: PC/Windows 10/ Chrome 84.0.4147.105

alexmojaki:
No it's fine, thanks for helping.

Issue #21, opened 2020-04-10
Bonus challenges


alexmojaki:
Sometimes we'd like to give users more difficult exercises which are optional and not part of the linear course flow. Currently neither the frontend nor the backend are equipped to handle this. I'd like to design a user interface that lets users complete challenges on the side. Any thoughts on how this should look?

PR #50, opened 2020-08-12
Get page/step from frontend, not database, allowing multiple open pages at once

### LINKED ISSUE #43: Get current page from the frontend instead of the database when running and checking code
The backend gets the current page from the user object in the database and uses that to check if the code passes the current step. This means that if the user has two pages open, trying to complete a step on one of them will fail. Instead the frontend JS should send through the current page index when running code.

alexmojaki:
Closes #43 

This changes a whole bunch of things so it needs testing badly. @spamegg1, @BobAnkh please play with the deployed environment and test it thoroughly if you can. Test running code, navigating between pages, moving steps in developer mode, hints, solutions, make sure that the editor/snoop/pythontutor are shown and hidden at the right times, open multiple tabs at different pages and make sure you can interact with them independently.

### DIFF
--- a/backend/main/chapters/c04_for_loops.py
+++ b/backend/main/chapters/c04_for_loops.py
@@ -624,16 +624,87 @@ b   b
                     print(inner_line)
                 print(line)
 
-    final_text = """
-Sweet! You're really getting the hang of this! If you want, here's one more optional bonus challenge. Try writing a program that outputs:
+    class diagonal_name_bonus_challenge(ExerciseStep):
+        """
+Sweet! You're really getting the hang of this!
+
+If you want you can do one more optional bonus challenge below.
+If not, you can just continue to the [next page](/course/?page=BasicTerminology) now.
+You can come back and do this later if you want.
+
+Try writing a program that outputs the given `name` in a diagonal line, e.g:
 
     W
      o
       r
        l
         d
+        """
 
-Or don't, it's up to you.
+        hints = """
+The first letter should have 0 spaces before it, the second letter should have 1 space before it, the third should have 2, etc.
+You should keep the spaces in a variable and build them up in a loop, as before.
+The difference is that you need to print letters at the same time as building up spaces.
+In other words, you need a single loop that does both.
+The body of the loop needs to print the spaces and letter, and also add a space.
+Since the first letter should have no spaces before it, you need to add a space after printing a letter.
+        """
+
+        # TODO automatically catch print with multiple args?
+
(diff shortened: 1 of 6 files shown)


spamegg1:
@alexmojaki Will do.

alexmojaki:
Thanks @spamegg1. In case it's not clear, I'm asking if you can test it out now, not after I merge. Opening this PR automatically deployed it to a new environment to test before making it live. Click "View Deployment to check it out. Let me know if you spot any problems before I merge. You can login with admin@example.com, password "admin".

I've just pushed another commit which gives a proper bonus challenge at the end of the BuildingUpStrings page. Before it was just described in the text, now it's a proper exercise with hints and stuff. At the same time it provides a link to easily skip to the next page. It's a bit of a hacky solution to #21 and nonlinear content in general but I think it's a good start. Let me know what you think.

spamegg1:
Ah OK, I misunderstood. Saw the View Deployment link, will check it out.

spamegg1:
Everything seems to be working fine. Used it on multiple tabs, etc.

The "Reveal" button at the end of a Hint sequence seems to do nothing. After clicking "Yes" on the "Are you sure..." question, it displays the solution and the Reveal button. If I click the Reveal button 25-30 times it disappears.

I'm still having the "Oops, you need to set word = 'Hello' before we can continue." issue.

BobAnkh:
I have also tested the deployment. 

> Test running code, navigating between pages, moving steps in developer mode, hints, solutions, make sure that the editor/snoop/pythontutor are shown and hidden at the right times, open multiple tabs at different pages and make sure you can interact with them independently

I have tested these things. Using Light Mode, so everything works fine to me, LOL.

Forget where the issue mentioned by @spamegg1 locates, so I haven't tested it.

alexmojaki:
Great, thanks for testing guys. Also see how I implemented the bonus challenge at the end of the BuildingUpStrings page. New possibilities are opening up!

> I'm still having the "Oops, you need to set word = 'Hello' before we can continue." issue.

What's the issue? You have to define the variable to use it. Did you define it and it disappeared a short while later?

spamegg1:
Yep, checked out the bonus after Building up Strings, very cool.

> Did you define it and it disappeared a short while later?

Yep. Might be some subtle browser/extension thing. Let's not worry about it.

alexmojaki:
No, the browser shouldn't have anything to do with it. See if you can figure out how to reproduce it.

PR #55, opened 2020-08-17
Added Chapter 08 Combining Booleans

spamegg1:


alexmojaki:
Thanks! This looks pretty good. I've given some feedback in https://github.com/spamegg1/futurecoder/pull/1. Most of it is actually stuff I mentioned before that you may have forgotten, plus I'm experimenting with some new ways of doing things. Let me know what you think. You can merge it if you like.

I still need to go through the rest of the chapter, so there's more to come.

Quick note: The first percentage exercise uses `or`, so it should be in the page about `or`, not the page about `and`.

spamegg1:
Thanks, merged.

I was on the fence about that page issue. First I put it at the end of the previous `or` page. Then I moved it to the `and` page because two exercises use the same `percentage` code. I'll just move it back.

alexmojaki:
Working through this PR has got me thinking that we need to teach defining functions before this chapter and then use that in this chapter. I delayed functions previously because for beginners they're not that interesting and come with technical complications. But that was relative to the other material like loops and if statements. Now I'm realising that boolean operators are also uninteresting and more complicated than I thought.

And I think this chapter would be much nicer with functions. In general once we have functions I want to use them to teach in a sort of TDD style. So the chapter might start with this code:

```python
def greeting(name):
    result = "Hello " + name + "!"
    if name == "Alice":
        result += "I have a friend called Alice!"
    if name == "Bob":
        result += "I have a friend called Bob!"


assert greeting("Alice") == "Hello Alice! I have a friend called Alice!"
assert greeting("Bob") == "Hello Bob! I have a friend called Bob!"
assert greeting("Charlie") == "Hello Charlie!"
```

or maybe even:

```python
def is_friend(name):
    if name == "Alice":
        return True
    elif name == "Bob":
        return True
    else:
        return False
```

with similar tests. When we reach the common mistake they can see the tests fail for Charlie. I like the idea of that better than "run this same code several times and type in different names". But maybe I'm wrong and the interactivity is good for keeping students engaged with the material. What do you think?

Sorry for suggesting such a big change after so much work, but I think the overall structure of the chapter won't be affected much by this idea.

spamegg1:
OK... so... I would have to assume that functions are covered before Chapter 08? Or introduce functions in this chapter?

Agreed, this would have been much better with functions. In fact I was struggling to write it without functions. 

Also, having functions covered would allow me to jump into Tic-Tac-Toe project after this. (I guess there is still `while` loops to cover)

alexmojaki:
Yes, this chapter would come after functions. In fact it would make sense to cover while loops and nested lists first as well to allow going straight to tic-tac-toe after this chapter.

spamegg1:
OK, I'll think about how to integrate functions into this chapter... Is there anything I need to know for testing functions? I don't believe there is any existing code that I could look at that does that. Does `ExerciseStep` testing work the same way? Do we make the user write the `assert` code or is that for testing?

alexmojaki:
No, the system isn't ready for that yet, it will need to be implemented. It'll actually be easier than the existing method though. Don't worry about the code yet, just think about what you'd like the user experience to look like.

> Do we make the user write the assert code or is that for testing?

For example, what do you think this should be like for the user?

spamegg1:
OK got it, won't worry about the code for now.

It would be nice if the user learned about `assert` and started using it for testing eventually, that's what I did in my project with the prime-finding code. But maybe it's a bit too early? I introduced it rather later there.

I'll think about it for a bit.

alexmojaki:
OK, I have added a chapter for functions. Please test it out. By the way, the website is now https://futurecoder.io/

The functions chapter is not fully complete but it has all the concepts needed to write this chapter on booleans, and the framework now supports exercises with real functions. I will write some documentation about it soon.

PR #60, opened 2020-09-05
Improvements to function-based exercises

alexmojaki:

Issue #35, opened 2020-08-04
Should we encourage or discourage copy/pasting code from the text into the editor/shell?


alexmojaki:
The text has a lot of steps that say "run this exact code". The easiest thing for the user to do is then select the code and copy and paste it.

The problem is that students learn and absorb much better if they manually type in the code (citation needed, too lazy right now). So I want to prevent or discourage copy/pasting.

On the other hand, students will get annoyed if they have to type in lots of code. Sometimes the code they have to run is quite long, and if they don't get it exactly right (except some small differences like spacing or quotes which are OK - the AST has to match) they won't proceed.

Possible courses of action:

1. Prevent copying from the text entirely. Or conversely, prevent pasting into the editor/shell.
2. Prevent copying certain snippets of code, particularly small ones.
3. Allow all copying, but the first time a student does it, show a pop up suggesting that this is bad for learning and they should try to avoid it.
4. Accept that students don't want to type code, and make it easier to run code without typing. This could be combined with either 2 (in cases that copying is allowed) or 3. We could provide a button that copies the code to the clipboard, or even a button that immediately runs the code in the shell/editor.

I prefer either 2 or 3. Thoughts? Ideas?

spamegg1:
This is a tricky balance... We don't want to annoy the user. 

I would be in favor of something in between 2 and 3: tell the student to "Type" explicitly in text, maybe disallow copying small snippets, allow copying medium snippets, but have a "Copy" button for very large snippets and discourage typing those. 

That is what I did in my project. Only the really large snippets have a "Copy" button, and only when the code in the large snippet has already been discussed earlier in smaller chunks, and the large snippet is given as a check to make sure the user did not make a mistake: "this is what your code should look like so far" type of thing.

If this is too complicated, I'm in favor of 2.

aninuth:
As a computer science student, one thing that I've seen professors do often is to not actually provide fully written out code.  They'll make pseudocode available and be very specific about the structure and intended function of the code, but the task of translating those ideas into the language that we're working in must be completed by the student.  This might be a little too advanced for beginners (especially if the goal is to teach syntax and not focusing on problem-solving), but I've found it to be rather helpful.

alexmojaki:
[...]

PR #63, opened 2020-09-07
Added c09_combining_booleans.py

spamegg1:


spamegg1:
Let's see how it goes!

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+We learned about *booleans* (`True` and `False`) when we introduced If statements.
+We also learned about comparison operators `==`, `<`, `>`, `<=`, `>=` which return booleans.
+Now we want to combine booleans to check for more complex conditions.
+
+Here's a simple example: imagine you have two friends, Alice and Bob.
+The function below accepts one parameter, `name`, and checks if the person with the given name is among your friends.
+Below its definition, we test the function on three different inputs with `assert_equal` from the previous chapter.
+After that, we have an example program that uses this function interactively.
+Type and run the code in the editor:
```
alexmojaki, line 21: As discussed in #35, I think this is an example of when we would write "copy" rather than "type". This snippet is long and doesn't have any unfamiliar concepts yet.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+Let's see if we can do better.
+We can combine the `if` and `elif` statements using **`or`**. `or` is a *boolean operator*,
+meaning it's an operator (like `+` or `-`) which combines two booleans (`True` or `False`).
+
+The expression `A or B` is `True` if either `A` or `B` is `True`, i.e. if `A` is `True` or `B` is `True`, or both.
+It's only `False` if neither `A` nor `B` is `True`, i.e. both are `False`.
+
+Let's try some examples in the shell.
+Try the following in the shell. Think about what you expect it to return:
```
alexmojaki, line 55: Bit of repetition here.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+                elif name == "Bob":
+                    return True
+                else:
+                    return False
+
+            assert_equal(is_friend("Alice"), True)
+            assert_equal(is_friend("Bob"), True)
+            assert_equal(is_friend("Charlie"), False)
+
+            # Interactive program
```
alexmojaki, line 38: Note that comments haven't been introduced yet

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+Take careful note of how we wrote the condition:
+
+    return name == "Alice" or name == "Bob"
+
+A common mistake is to write this instead:
+
+    return name == "Alice" or "Bob"
+
+It makes sense if you read it like English: "If `name` is equal to either "Alice" or "Bob", then return True...".
```
alexmojaki, line 157:
```suggestion
It makes sense if you read it like English: "If `name` is equal to either `"Alice"` or `"Bob"`, then return True...".
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+                return name == "Alice" or "Bob"
+
+            assert_equal(is_friend("Alice"), True)
+            assert_equal(is_friend("Bob"), True)
+            assert_equal(is_friend("Charlie"), False)
+
+    class InspectWithBirdseye(ACommonMistake):
+        text = """
+The second and third tests fail! Our function seems to be doing the wrong thing:
+it returns `Bob` (a string, not a boolean!) when `name` is `Charlie`. What is going on?
```
alexmojaki, line 175:
```suggestion
it returns `"Bob"` (a string, not a boolean!) when `name` is `"Charlie"`. What is going on?
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+Try inspecting the code with Birdseye. Inspect the `return` statements of each `is_friend` calls carefully.
```
alexmojaki, line 176:
```suggestion
Try inspecting the code with Birdseye. Inspect the `return` statements of each `is_friend` call carefully.
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+(Use the blue arrow buttons)
+        """
+
+        def check(self):
+            return super().check() and self.code_source == "birdseye"
```
alexmojaki, line 181:
```suggestion
        expected_code_source = "birdseye"
```

New feature!

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+
+    class AnExercise(ExerciseStep):
+        """
+When we inspect it with Birdseye, we can see that:
```
alexmojaki, line 185:
```suggestion
When we inspect it with Bird's Eye, we can see that:
```

I know I wrote this part, but I've since decided to spell it like that within futurecoder. I feel weird about it but the one word spelling is good for `import birdseye` and `pip install birdseye`, not this.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+
+Python actually lets you treat anything like a boolean. Most things are equivalent to `True` in this context,
+including all strings (such as `"Bob"`) except the empty string which is 'falsy'.
+
+> Is there a better way to write the condition without repeating `name ==` each time?
+
+Yes! In [Functions and Methods for Lists](/course/?page=FunctionsAndMethodsForLists) we mentioned the `in`
+operator, which you can use with a list like this:
+
+    if name in ["Alice", "Bob", "Charlie"]:
```
alexmojaki, line 235:
```suggestion
    return name in ["Alice", "Bob", "Charlie"]:
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+            return False
+            
+As before, we can simplify this solution to:
+
+    def is_valid_percentage(x):
+        return 0 <= x and x <= 100
+
+In Python,
+
+    if 0 <= x and x <= 100:
```
alexmojaki, line 363:
```suggestion
    0 <= x and x <= 100
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+
+can also be written as a short hand:
+
+    if 0 <= x <= 100:
```
alexmojaki, line 367:
```suggestion
    0 <= x and x <= 100

can also be written as a short hand:

    0 <= x <= 100
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+                """
+
+        def solution():
+            @returns_stdout
+            def is_winning_row(_, row: List[str]):
+                if row[0] == row[1] and row[0] == row[2]:
+                    print("We have a winner!")
+                else:
+                    print("Not a winning row.")
+            return is_winning_row
```
alexmojaki, line 406:
```suggestion
        def solution(self):
            def is_winning_row(row: List[str]):
                if row[0] == row[1] and row[0] == row[2]:
                    print("We have a winner!")
                else:
                    print("Not a winning row.")
            return is_winning_row
```

The API changed in #60.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+                    (row2[0] == row2[1] and row2[0] == row2[2]) or
+                    (row3[0] == row3[1] and row3[0] == row3[2])
+                ):
+                    print("We have a winner!")
+
+            row1 = ["X", "O", "X"]
+            row2 = ["X", "X", "X"]
+            row3 = ["O", "O", "X"]
+
+            is_winning_board(row1, row2, row3)
```
alexmojaki, line 531: Now that we have functions, this has become a bad example. We're showing them bad code. This should be written as:

```python
def is_winning_row(row):
    return row[0] == row[1] and ...

def is_winning_board(row1, row2, row3):
    return is_winning_row(row1) or ...
```

I think we need to think of a new example.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+            (middle == row1[2] and middle == row3[0])
+        ):
+            print(middle + " wins!")
+        else:
+            print("No winners.")
+
+Unlike the other two boolean operators `and` and `or`,
+which are used in between two booleans (called *binary* operators),
+`not` is used before only one boolean (called a *unary* operator).
+It negates the statement to which it is applied. Try in the shell:
```
alexmojaki, line 599:
```suggestion
It negates the expression to which it is applied, a bit like a minus sign. Try in the shell:
```

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+"""),
+            ((["X", "O", "O"], ["X", "O", "X"], ["O", "X", "X"]), """\
+O wins!
+"""),
+            ((["X", "O", "X"], ["X", "O", "X"], ["O", "O", "X"]), """\
+No winners.
+"""),
+        ]
+
+    class IntroducingNot(VerbatimStep):
```
alexmojaki, line 582: New page.

backend/main/chapters/c09_combining_booleans.py
```diff
@@ -0,0 +1,638 @@
...
+Similar to how we checked a row, check that the 3 entries on a diagonal are equal to each other, using `and`.
+Similar to how we checked three rows together, check the two diagonals together, using `or`.
+As before, use a multi-line boolean expression where `if` contains lines that are connected with `or`. Only 2 lines 
+needed.
+Don't forget the `else` part!
+        """
+
+        def solution():
+            @returns_stdout
+            def diagonal_winner(_, row1: List[str], row2: List[str], row3: List[str]):
```
alexmojaki, line 559: Type annotations aren't enough here, you will need to override `generate_inputs`.

alexmojaki (changes requested):
Well done! This is looking good. Most of the comments are minor. But we do both need to think about good examples for `not` and combining the operators.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Got it. Changing "type" to "copy".

backend/main/chapters/c09_combining_booleans.py
spamegg1: Removed.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Oh really? Didn't know that. Removed.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Right! I was just using the previous version you wrote for me.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Got it.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Right, this is left over from the previous incarnation, must have missed it.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Same issue, if became return...

backend/main/chapters/c09_combining_booleans.py
spamegg1: Got it.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Done

backend/main/chapters/c09_combining_booleans.py
spamegg1: One easy cheat is to change the exercise to find winning Columns instead.

backend/main/chapters/c09_combining_booleans.py
spamegg1: Would this work:
```python
        def generate_inputs(cls):
            return {
                "row1": [random.choice(["X", "O"]) for _ in range(3)],
                "row2": [random.choice(["X", "O"]) for _ in range(3)],
                "row3": [random.choice(["X", "O"]) for _ in range(3)]
            }
```

spamegg1:
One simple `not` exercise is to ask them to simplify the pattern:
```python
if x:
    return False
else:
    return True
```
into `return not x`, in the context of the `or` solution to `is_valid_percentage`. So the solution looks like:
```python
def is_valid_percentage(x):
    return not (x < 0 or x > 100)
```
At the end we can mention how this is equivalent to the solution using `and`: 
`not (x < 0 or x > 100)` is equivalent to
`not (x < 0) and not (x > 100)` is equivalent to
`x >= 0 and x <= 100`

backend/main/chapters/c09_combining_booleans.py
alexmojaki: I think so, and yes to `@classmethod`.

To see this in action, try cheating by submitting a solution that passes the manual `tests` just using hardcoded if statements. See the difference with and without `generate_inputs`.

spamegg1: Oh I see. Random testing is necessary. I'll make the changes and commit this.

backend/main/chapters/c09_combining_booleans.py
alexmojaki: No. Eventually we will have introduce nested lists before this chapter and the examples will be modified to reflect that instead of having three lists. Then it'll be `def is_winning_column(board, column): return board[0][column] == ...`.

I think it's actually fine to skip to the exercise with diagonals. We've shown them the main pitfall and they have plenty of tools to help them. I no longer think we need to be so thorough, this isn't like loops.

spamegg1: OK! That makes my life easier! Will make the changes and commit.

alexmojaki:
I think you have the right general idea but I think it's stretching the percentage thing too far. Here's a proposal. You can still suggest something else if you like.

```
Suppose you're writing a program which processes images. Only certain types of file can be processed.
If the user gives you a file that can't be processed, you want to show an error:

    if invalid_image(filename):
        print("I can't process " + filename)

Here's an example function to do that:

    def invalid_image(filename):
        if filename.endswith(".png") or filename.endswith(".jpg"):
            return False
        else:
            return True

    assert_equal(invalid_image("dog.png"), False)
    assert_equal(invalid_image("cat.jpg"), False)
    assert_equal(invalid_image("invoice.pdf"), True)

This is longer than it needs to be. Rewrite `invalid_image` so that the body is a single line `return <expression>`,
i.e. no `if` statement.

----

Well done! Here are two valid solutions:

    def invalid_image(filename):
        return not (filename.endswith(".png") or filename.endswith(".jpg"))

    def invalid_image(filename):
        return filename.endswith(".png") and not filename.endswith(".jpg")

(if you're curious, these are equivalent because of something called De Morgan's law)
```

spamegg1:
This makes sense, it will force them to use `not` unlike my example. Also it's different. I was thinking some example with electrical circuits and logic gates but it got too complicated. I'll add this in.

alexmojaki:
Feedback here: https://github.com/spamegg1/futurecoder/pull/1

When you make a new repo or fork make sure you watch it.

spamegg1:
Thanks, I was not aware of the Watch feature. Still learning Github!

alexmojaki:
Feedback regarding `not`: https://github.com/spamegg1/futurecoder/pull/2

Yeah I haven't seen anyone else make PRs like this, but I don't see any reason why not. It's good for big feedback and collaboration. You should have a chance to review my work and offer your opinion when I'm making changes to code you wrote!

Anyway, after this I think we can merge. Then I'm going to create some conveniences for disallowing certain syntax so we can resolve many of the TODOs without writing a bunch of manual MessageSteps because this pattern appears a lot.

spamegg1:
I've reviewed the changes, they look good to me. I tend to be terse, so you balance me out quite well.

I can see that you are moving away from writing functions with `print` outputs. That's good! I was simply following your earlier drafts.

It's my fault I didn't realize the need for `generate_inputs` on the image file exercise. I thought that since it's just `str` type hints would be enough.

Do you have to manually change `backend/main/test_transcript.json` every time new content is added? Was that part of my job that I didn't do? Or is the build supposed to automatically generate that stuff?

> we can resolve many of the TODOs without writing a bunch of manual MessageSteps because this pattern appears a lot.

Ah, the pleasures of metaprogramming :100:

alexmojaki:
I like to be terse as well. It's a very tricky balance trying to decide how complicated/important a concept is and thus how much explaining is needed.

See https://github.com/alexmojaki/futurecoder/blob/master/how_to_contribute.md#testing .

Once the review app has finished deploying, go through the whole chapter. Look at the hints and solutions. Follow the instructions like "modify this line".

alexmojaki:
> One thing is that invalid_image solution works by hitting Run without rewriting it. Is this a problem? Probably OK, user knows it's already a correct solution.

It's in the TODOs.

alexmojaki (approved):

alexmojaki:
OK, that's it! Congratulations on being the first person to make a major contribution to this project! Thanks a lot for doing this, I know it was a long and complicated process but it was very helpful for me and I'm sure you learned a lot. I think this will get a lot easier. Hopefully this is the start of great things!

spamegg1:
YAY! :partying_face: 
Thanks a lot for the effort you put in! It was indeed long and complicated.
Yep certainly learned a lot, especially with the API and Github itself.

alexmojaki:
TODOs resolved in #64, particularly https://github.com/alexmojaki/futurecoder/pull/64/files#diff-ca87fa8c7d507e8d52adcd3442ad7409. Maybe try out the deployment and see if you can 'cheat'.

spamegg1:
Excellent! Can't cheat anymore, also the syntax highlighting is working! Looks much more beautiful.

Issue #66, opened 2020-09-13
Make bonus challenges concrete


alexmojaki:
There are currently three bonus challenges in the text (search the chapters for 'bonus'). The first one is in chapter 4 and is a proper exercise with hints and a solution. The other two don't have this, they're just described in the text. They should be converted to real ExerciseSteps and moved to the end of the page with a link to the next page, just like the first one.

@spamegg1 this is a decent bite sized task, want to give it a go?

spamegg1:
@alexmojaki Yeah sure! I was planning to tackle #34 but I'll give this a go. That one would be much longer.

spamegg1:
The solution to the `string_sum` bonus is quite tricky: not adding the separator after the last word. They don't know list slicing, or string slicing, or `range` at this point do they?

Without those I can only come up with this somewhat hacky solution:
```python
def solution(self, words: List[str]):
    separator = ' - '
    total = ''
    total_with_separator = ''

    for word in words:
        total = total_with_separator + word
        total_with_separator = total + separator
    print(total)
```
I can explain this with hints, just making sure I'm not missing something obvious.

alexmojaki:
What about with a boolean flag and an if statement?

spamegg1:
OK... something like this:
```python
        def solution(self, words: List[str]):
            separator = ' - '
            total = ''
            not_finished = True

            for word in words: 
                total += word
                if not_finished:
                    total += separator

            print(total)
```

alexmojaki:
It is tricky, which is why it's a bonus. Fortunately there's a recent exercise with some similarity.

```python
from typing import List


def solution(self, words: List[str], separator: str):
    total = ''
    not_first = False

    for word in words:
        if not_first:
            total += separator
        total += word
        not_first = True

    print(total)


solution(None, ["a", "b", "c"]," - ")
```

spamegg1:
Ah, add the separator first... makes sense.

spamegg1:
Ah... this is going to go horribly.
`docker-compose build` and `docker-compose up` run fine.
But the page at `localhost:3000/course` gives me:
> Method: load_data
Request data: {}
Traceback (most recent call last):
  File "/usr/src/app/backend/main/views.py", line 51, in api_view
    result = method(**args)
  File "/usr/src/app/backend/main/views.py", line 155, in load_data
    page_index=pages[self.user.page_slug].index,
KeyError: 'CombiningAndWithOr'

[...]

PR #67, opened 2020-09-14
Made bonus challenges concrete

spamegg1:
addressing #66

### DIFF
--- a/backend/main/chapters/c06_lists.py
+++ b/backend/main/chapters/c06_lists.py
@@ -93,9 +93,9 @@ Is there a similar concept among strings to 0? A blank initial value?
             (['The', 'quick', 'brown', 'fox', 'jumps'], 'Thequickbrownfoxjumps'),
         ]
 
-    class double_numbers(ExerciseStep):
+    class strings_sum_bonus(ExerciseStep):
         """
-Optional bonus challenge: extend the program to insert a separator string *between* each word.
+Optional Bonus challenge: extend the program to insert a separator string *between* each word.
 For example, given
 
     words = ['This', 'is', 'a', 'list']
@@ -105,6 +105,62 @@ it would output:
 
     This - is - a - list
 
+        """
+        hints = """
+This is similar to the previous exercise. You can start with your solution from that.
+You need to modify the string that you are adding at each step.
+In addition to each word in the list, you also have to add the separator.
+But you don't want to add the separator after adding the last word in the list.
+Unfortunately there is no "subtraction" with strings; you can't add the last separator then remove it.
+This is tricky! Using a for loop, how can you perform an operation at all steps except the last? Well, you can't (for now).
+What if you reversed it? Add the separator first, add the word second?
+This way, you have to add the separator at each step except the FIRST step.
+The solution is somewhat similar to the program from the "Understanding Programs with Snoop" section.
+This can be done with a for loop (with an if statement inside) and a boolean variable (to keep track of whether we 
+are at the first step or not).
+Define a boolean variable BEFORE the for loop. Then change its value INSIDE the loop (after adding 
+the necessary strings).
+        """
+
+        # TODO: MessageStep: catch the "obvious solution" where the user adds the separator after the last word?
+        # TODO: is overriding generate_inputs needed?
+
+        def solution(self, words: List[str], separator: str):
+            total = ''
+            not_first = False
+
+            for word in words:
+                if not_first:
+                    total += separator
+                total += word
+                not_first = True
+
+            print(total)
+
+        tests = [
+            ((['This', 'is', 'a', 'list'], ' - '), 'This - is - a - list'),
+            ((['The', 'quick', 'brown', 'fox', 'jumps'], ' - '), 'The - quick - brown - fox - jumps'),
+        ]
+
+    final_text = """
+Congratulations! That was very tricky! One solution looks like this:
+
+    total = ''
+    not_first = False
+
+    for word in words:
+        if not_first:
+            total += ' - '
+        total += word
+        not_first = True
+
+    print(total)
+        """
+
+
+class BuildingNewLists(Page):
+    class double_numbers(ExerciseStep):
+        """
 Lists and strings have a lot in common.
 For example, you can add two lists to combine them together into a new list.
 You can also create an empty list that has no elements.
@@ -529,8 +585,8 @@ output:
     o r
     d l
     b d
-    y  
-    e  
+    y
+    e
 
 and for:
 
@@ -755,7 +811,7 @@ Let's review how to work with lists. Suppose we have a list `nums = [1, 2, 3]`. 
 Here's some new things. Try them out in the shell.
 
 - **`subscript assignment`**: Set a value at an index. `nums[0] = 9` changes the list to `[9, 2, 3]`.
-- **`join`**: Add a list of strings with a separator in between. This is a method of strings (the separator) which takes an iterable of strings as an argument. `'--'.join(['apples', 'oranges', 'bananas'])` returns `'apples--oranges--bananas'`. You can also use an empty string if you don't want a separator, e.g. `''.join(['apples', 'oranges', 'bananas'])` returns `'applesorangesbananas'`. 
+- **`join`**: Add a list of strings with a separator in between. This is a method of strings (the separator) which takes an iterable of strings as an argument. `'--'.join(['apples', 'oranges', 'bananas'])` returns `'apples--oranges--bananas'`. You can also use an empty string if you don't want a separator, e.g. `''.join(['apples', 'oranges', 'bananas'])` returns `'applesorangesbananas'`.
 - **`sum`**: Add a list of numbers. `sum(nums)` is 6.
 - **`in`**: A comparison operator that checks if a value is in a list. `2 in nums` is `True`, but `4 in nums` is `False`.
 - **`index`**: Returns the first index of a value in a list. `[7, 8, 9, 8].index(8)` is 1. Raises an error if the value isn't there.
@@ -818,7 +874,7 @@ But suppose you don't want the 9 to be at the end, you want it to go between the
 
     [1, 2, 9, 3, 4, 5]
 
-Replace the middle line (i.e. the call to `append`) with the right function/method call to do that. 
+Replace the middle line (i.e. the call to `append`) with the right function/method call to do that.
         """
 
         hints = """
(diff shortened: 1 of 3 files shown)


spamegg1:
I already see a few newline mistakes I made with the hints in Chapter 6.

backend/main/chapters/c06_lists.py
```diff
@@ -105,6 +105,62 @@ class double_numbers(ExerciseStep):
...
+This is tricky! Using a for loop, how can you perform an operation at all steps except the last? Well, you can't (for now).
+What if you reversed it? Add the separator first, add the word second?
+This way, you have to add the separator at each step except the FIRST step.
+The solution is somewhat similar to the program from the "Understanding Programs with Snoop" section.
+This can be done with a for loop (with an if statement inside) and a boolean variable (to keep track of whether we are at the first step or not).
+Define a boolean variable BEFORE the for loop. Then change its value INSIDE the loop (after adding the necessary strings).
+        """
+
+        # TODO: MessageStep: catch the "obvious solution" where the user adds the separator after the last word?
+        # TODO: is overriding generate_inputs needed?
```
alexmojaki, line 124: No, since the problem is the same for any random strings

backend/main/chapters/c06_lists.py
```diff
@@ -105,6 +105,62 @@ class double_numbers(ExerciseStep):
...
+                if not_first:
+                    total += separator
+                total += word
+                not_first = True
+
+            print(total)
+
+        tests = [
+            ((['This', 'is', 'a', 'list'], ' - '), 'This - is - a - list'),
+            ((['The', 'quick', 'brown', 'fox', 'jumps'], ' - '), 'The - quick - brown - fox - jumps'),
```
alexmojaki, line 140: Test two different separators

backend/main/chapters/c06_lists.py
spamegg1: OK. That's what I thought.

backend/main/chapters/c06_lists.py
spamegg1: Makes sense...

backend/main/chapters/c07_nested_loops.py
```diff
@@ -399,14 +399,62 @@ def solution(self, players: List[str]):
...
+        for j in range(len(players)):
+            if i < j:
+                print(players[i], 'vs', players[j])
+
+For a more elegant solution, you can change the second `range` expression, so that
+the `if` condition is automatically satisfied and we avoid unnecessary computation:
+
+    players = ["Alice", "Bob", "Charlie"]
+    for i in range(len(players)):
+        for j in range(i + 1, len(players)):
```
alexmojaki, line 450: I agree with your general idea but multiple arguments to `range` haven't been covered.

backend/main/chapters/c07_nested_loops.py
```diff
@@ -399,14 +399,62 @@ def solution(self, players: List[str]):
...
     Alice vs Bob
     Alice vs Charlie
     Bob vs Charlie
 
+        """
+
+        def solution(self, players: List[str]):
+            for i in range(len(players)):
+                for j in range(i + 1, len(players)):
+                    print(players[i], 'vs', players[j])
```
alexmojaki, line 415: This solution is shown to users once they get stuck, so I generally prefer it to be a more basic solution like the first one you show.

Another possible solution is to just change `!=` to `<` in the original solution:

```python
            for player1 in players:
                for player2 in players:
                    if player1 < player2:
                        print(player1, 'vs', player2)
```

backend/main/chapters/c07_nested_loops.py
alexmojaki: (and this isn't the place to explain it)

backend/main/chapters/c07_nested_loops.py
```diff
@@ -399,14 +399,62 @@ def solution(self, players: List[str]):
...
+
+        hints = """
+This is similar to the previous exercise, you'll need a for loop inside a for loop like before.
+This time, in your loops you need to use `range` and `len`, and use indexing to access the list entries.
+Look at the desired output: `Alice vs Bob`: `Alice` comes before `Bob` in the `players` list.
+Similarly in `Alice vs Charlie` `Alice` comes before `Charlie`, and in `Bob vs Charlie` `Bob` comes before `Charlie` in `players`.
+We don't want to print `Bob vs Alice` because `Bob` comes AFTER `Alice` in `players`.
+The only pairs we want to print are those where the left player comes before the right player in the list.
```
alexmojaki, line 423: Ah, this is actually one of the disadvantages of a concrete exercise that needs to check that the output is exactly right (as opposed to letting the user decide for themselves in a soft bonus challenge). There's only one correct output. I hadn't thought about it here. The user might feel that `Bob vs Alice` is acceptable instead of `Alice vs Bob`. We need to specify the correct output precisely and clearly in the text, it can't only be hidden in hints. Ideally we want users to try and solve problems without looking at the hints, and they should always know exactly what the goal is.

backend/main/chapters/c07_nested_loops.py
```diff
@@ -399,14 +399,62 @@ def solution(self, players: List[str]):
...
+How can we express this relation in terms of the list indexes of the two for-loops?
+Once you figure out the relation, you can express it with an `if` statement, or inside the second `range` expression.
+                """
+
+        tests = [
+            (["Alice", "Bob", "Charlie"], """\
```
alexmojaki, line 429: Generally at least two tests.

backend/main/chapters/c07_nested_loops.py
spamegg1: OK

backend/main/chapters/c07_nested_loops.py
spamegg1: OK, I'll replace it with the basic one.

backend/main/chapters/c07_nested_loops.py
spamegg1: I thought about that, but wasn't sure how to go about it. Wanted to hear your take. Another potential issue is that since the names in the list are already in alphabetical order (but they don't have to be) the user might think that they only need to print pairs in alphabetical order (rather than list order). Maybe ruin the alphabetical order in the list, and ask the question that way? (then specifying the only pairs to be printed are the ones left-to-right on the list)

backend/main/chapters/c07_nested_loops.py
spamegg1: OK, will add.

backend/main/chapters/c07_nested_loops.py
spamegg1: Removed this for now.

backend/main/chapters/c07_nested_loops.py
alexmojaki: Yes, ruin the alphabetical order. Good idea.

backend/main/chapters/c06_lists.py
```diff
@@ -93,9 +93,9 @@ def solution(self, words: List[str]):
             (['The', 'quick', 'brown', 'fox', 'jumps'], 'Thequickbrownfoxjumps'),
         ]
 
-    class double_numbers(ExerciseStep):
+    class strings_sum_bonus(ExerciseStep):
         """
-Optional bonus challenge: extend the program to insert a separator string *between* each word.
+Optional Bonus challenge: extend the program to insert a separator string *between* each word.
```
alexmojaki, line 98: To make an exercise optional, users need to be able to skip past it easily. Don't make them open the table of contents, provide a link to the next page. See the existing bonus challenge.

backend/main/chapters/c06_lists.py
```diff
@@ -105,6 +105,61 @@ class double_numbers(ExerciseStep):
...
+Unfortunately there is no "subtraction" with strings; you can't add the last separator then remove it.
+This is tricky! Using a for loop, how can you perform an operation at all steps except the last? Well, you can't (for now).
+What if you reversed it? Add the separator first, add the word second?
+This way, you have to add the separator at each step except the FIRST step.
+The solution is somewhat similar to the program from the "Understanding Programs with Snoop" section.
+This can be done with a for loop (with an if statement inside) and a boolean variable (to keep track of whether we are at the first step or not).
+Define a boolean variable BEFORE the for loop. Then change its value INSIDE the loop (after adding the necessary strings).
+        """
+
+        # TODO: MessageStep: catch the "obvious solution" where the user adds the separator after the last word?
```
alexmojaki, line 123: It sounds like you're saying they submitted a solution to the previous exercise. I don't see why we would specially catch that. But maybe the description of the task should clarify the difference further.

backend/main/chapters/c07_nested_loops.py
```diff
@@ -399,14 +399,62 @@ def solution(self, players: List[str]):
...
+ 
+    players = ["Alice", "Bob", "Charlie"]
+    for i in range(len(players)):
+        for j in range(len(players)):
+            if i < j:
+                print(players[i], 'vs', players[j])
+    """
+
+
+class Exercises(Page):
```
alexmojaki, line 455: Page names need to be unique and descriptive on their own, not just in context of the chapter.

In any case I think this particular bonus exercise is sufficiently different from the original that it doesn't need to come right after it. Leave everything on one page and move the bonus exercise to the end. Then add a link to the next page.

backend/main/chapters/c07_nested_loops.py
spamegg1: Will do!

backend/main/chapters/c06_lists.py
spamegg1: Ah no... I meant the solution
```python
for word in words:
    total += word + separator
```
which would result in `This-is-a-list-`. Maybe this would not be a common mistake as I thought?

alexmojaki: Oh right, sorry, I got what you said but I'd forgotten what the previous exercise was and was thinking of an older exercise. Yes, a message for this is a good idea.

backend/main/chapters/c07_nested_loops.py
spamegg1: OK... so the bonus comes after the cracking password and upside down triangle exercises, to finish the page.

backend/main/chapters/c06_lists.py
spamegg1: OK, added links to both.

alexmojaki:
Nice! Thanks again!

PR #68, opened 2020-09-18
Discourage copying most code, add convenient copy button for big snippets

### LINKED ISSUE #35: Should we encourage or discourage copy/pasting code from the text into the editor/shell?

alexmojaki:
Closes #35

### DIFF
--- a/backend/main/chapters/c03_variables.py
+++ b/backend/main/chapters/c03_variables.py
@@ -185,7 +185,7 @@ class WritingPrograms(Page):
         """
 It's time to stop doing everything in the shell. In the top right you can see the *editor*. This is a place where you can write and run longer programs. The shell is great and you should keep using it to explore, but the editor is where real programs live.
 
-Copy the program below into the editor, then click the 'Run' button:
+Type the program below into the editor, then click the 'Run' button:
 
     __program_indented__
         """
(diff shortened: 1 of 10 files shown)


spamegg1:
Looks like a new addition to the API with `__copyable__`. Nice!
So I'm guessing that from now on, whenever I want to discourage copying, I'll have to exclude `__copyable__`.
Checked out the deployment, went through the course.
I got the `STOP!` warning the first time I tried to copy. Kinda freaked me out! Like "how is he doing this???" :)
The Copy clipboard buttons are working fine! But it's not too detrimental, because there are only few of them with the large codes.
Didn't run into any bugs. Great work as usual.

alexmojaki:
Thanks!

PR #72, opened 2020-09-29
Overhauled home page

ChrisjanWust:
WIP. Needs heavy clean-up, but functionality is there. Will follow up with some refactoring.


spamegg1:
Checked out the deployment. All the logos are gigantic for some reason, bigger than my screen (on Firefox). Tried it on Opera (chromium based) with the same results.
scr01
scr02
scr03

spamegg1:
Deployment looks good :+1:

ChrisjanWust:
Yeah, finally! Looks like a rebase did the trick.

spamegg1:
Deployment :+1: 
Maybe make the "Let's start" button green? 
(It still has to stand out from the background color)
And call it "Start Learning" or "Start the course"?
It would also help if it was closer to the center of the page. Grabs attention more easily.
I had trouble finding it at first. Kept scrolling to find the "Start" button.

alexmojaki:
I like that the colour of the get started button is original. There are green buttons all over the internet and all over futurecoder. But that's not a great justification.

I think the traditional text would be "Get started", and I don't care about that being original. But pretty much any text is fine.

I think the button should be bigger, both in general and relative to the Github button. Right now the Github button looks like the primary button, which may be the reason for @spamegg1's confusion.

@ChrisjanWust please change the About text to the earlier mentioned bullet points:

> futurecoder's goals are to:
> - Revolutionise computing education
> - Make the best learning resource possible that everyone can contribute to and improve
> - Allow anyone to learn programming, regardless of background or talent

I would very much like to see this merged soon. There is still much more to be added to this page and we can continue to refine details like the above afterwards but the main priority is to get something done and get rid of the current boring page.

ChrisjanWust:
Cool, I'll go through these small things ASAP.

1. Button text: what about "Start coding" ?
2. Colour: I'd very much like to keep the colour different to the rest of the page - not blue. I was going yellow to stick to python and contrast the blue. But green could work really well too, links well with "start" and is used in rest of future coder as well. Let me know if this needs more discussion, else I'm going ahead.
3. I disagree that the Github button looks like the main button. The colour that contrasts with the rest of the page is the yellow (or soon to be lime green).
4. I'll think about how to differentiate the two buttons more. Having the one bigger than the other would make for inconsistent styling. If this is a real problem, I'd rather consider moving the github button to another part of the page.
5. Will change the text of the about to goals.

alexmojaki:
It didn't occur to me that those were the Python colours. So maybe that doesn't work.

Overall, follow your heart.

PR #73, opened 2020-09-30
Make users predict the output of certain steps

alexmojaki:
The idea here is to avoid users mindlessly copying/transcribing code and running it. This pushes users to actually think about what code is doing. I got the idea from https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1006023#sec005

Please try this out and get a feel for the user experience. I have a lot of doubts about the design and need feedback. An easy place to try it out is the IntroducingVariables page. Turn on developer mode to easily go back and forth.

There are still a few things for me to do, such as annotating more steps in later chapters and disabling the editor and run buttons while the system is waiting for an answer.

A big change I want to make is what happens when the user gets the wrong answer. Instead of showing the correct answer, I think it should just eliminate the selected wrong answer and let the user try again with the remaining answers. This would continue repeatedly until the user gets the right answer or there is only one answer left. Does that sound good?


spamegg1:
> This would continue repeatedly until the user gets the right answer or there is only one answer left. Does that sound good?

I think it might be better to limit the number of attempts to 2-3, then show the correct answer.

alexmojaki:
> We want them to think through it carefully, not click away.

Hmmmm, but does your idea accomplish this? If someone already wants to get the correct answer at all I imagine they'd want to get it correct the first time and would already think carefully. Getting a wrong answer should still sting even if they get another chance. Is that just me?

I'm worried about the opposite problem. Some people may not want to deal with these exercises and may just click on answers until it goes away. My hope is that if this brute force method is tedious enough they might try to figure out the correct answer instead. But if they only get a few 'attempts' then brute forcing is easier.

spamegg1:
Hmm... this is very tricky indeed. Honestly I don't know the solution. 

Limiting the number of attempts is something a lot of MOOCs do, and I'm sure they have some data to back up its effectiveness.

I think that's just you. We are living in the age of instant gratification and millisecond long attention spans. (Part of the reason is the oceans of information overflow on the web and the need to sift through it quickly.) People these days will do everything to avoid even the slightest discomfort. You have no idea how many hundreds of times I had to explain the same super obvious things to people that are right there in front of them in big capital letters. They never looked or read even short sentences. 

I would assume that the length of careful, attentive thinking and reading would increase with one's coding experience and how much time and effort they have sunk into learning. Considering the fact that futurecoder is aiming at possibly non-programmers, we might end up scraping the bottom of the barrel in terms of attentive users. We get a lot of such people on Discord due to the pandemic and the "learn to code" meme that's been going around.

But I won't push this issue.

ChrisjanWust:
Really really like the feature! Successfully eliminates the mindlessness for me, similar to what the Brilliant app does.

Regarding the amount of attempts: I agree with @spamegg1. Having users retry till they've found the right answer is asking a lot.  And I say this, knowing that this course is aiming to challenge the status quo. Not only has everyone been getting used to answers being available everywhere, but we can't teach the unteachables. A student has to have a substantial amount of motivation / discipline to learn programming, and I feel we can't design for students who don't have at least 50% of what is required.

Furthermore, I think beginners genuinely might not know the answers to these questions at all. For example, a person who hasn't coded might have no clue what the output of `>>> word` would be. It could often happen that someone doesn't guess the right answer in the top 3 (or 4, for that matter), which I imagine the student would find incredibly discouraging.

I would even argue that attempts should be limit to either 1 or 2. I believe the effect of adding the feature is guaranteed to be positive if there's only one attempt, but it might just be very discouraging if there's an endless amount of attempts.

image

alexmojaki:
Great, thanks for your input @ChrisjanWust. I will give users a total of 2 attempts for every question.

> Furthermore, I think beginners genuinely might not know the answers to these questions at all. For example, a person who hasn't coded might have no clue what the output of `>>> word` would be.

This particular step is of course the first one I annotated, and I recognise it was a bad choice. I will remove the question. In general users should reasonably be able to infer the answer from the information given, although I'm stretching that when the answer is 'Error'.

alexmojaki:
OK, this is ready! @spamegg1 @ChrisjanWust please test one more time.

spamegg1:
Looking good... :+1:

alexmojaki:
Thanks for the reviews!

ChrisjanWust:
Yeah, I'm a bit late to the party, but very nice. Believe this feature really helps with the goal of getting students to not just click-n-copy through.

Issue #31, opened 2020-08-03
[enhancement] Suggestions for UI design and TOC


BobAnkh:
I'm new to this project so please forgive me if I get something wrong. 

Here is just my suggestions for the UI design and TOC after experiencing several chapters.

1. Is it possible to place the shell on the right or upper right in the first few chapters? Placing the shell in the lower right part with the upper right part all blank is a little bit uncoordinated to me.
2. Is it possible to place the TOC(table of content) first and let users to choose the part to learn ? I think different users may have different foundations and knowledge for python. And they can directly jump to the part they are more concerned about and bypass the basic parts they have already mastered that will surely cost them some time

Besides, is it more convinient for users to learn with it if we have translations?

Thank you for considering my suggestions.

alexmojaki:
Thanks for the feedback!

1. Someone else has mentioned this before, so I think it's worth doing. The shell can fill the right half of the screen until the editor appears. I invite you or anyone else to make a PR making this change.
2. This has also come up in discussions, and you're right, it should be easy to jump to any section. I will make a new issue soon detailing my thoughts on this.
3. Translations are not a priority at the moment. We don't even have one complete course, and until it's finished it's probably going to change.

BobAnkh:
Thank you for your reply!

I'm familiar with python but not with javascript. So maybe I can't help with this frontend issue. I will keep an eye on the project to see if there is something that I can help. I'm interested in this project.

If there is something that I can help with, please tell me.

alexmojaki:
The easiest thing anyone can help with is to brainstorm course content in #23.

spamegg1:
For the Table of Contents: it would be nice to have a topic tracker either at the top or bottom, displaying the name of the current topic, with buttons to other topics (similar to edX or Coursera).

alexmojaki:
@spamegg1 can you describe your idea in more detail?

alexmojaki:
@BobAnkh @spamegg1 I have implemented point 2 in #38. Please see the description and try out the test environment.

BobAnkh:
@alexmojaki When I try the page `Functions And Methods For Lists` in ch06, I suddenly got some ideas. But I'm not sure if this is what we really need, so I would like to discuss here first.

[...]

PR #74, opened 2020-10-02
improving lists chapter (fixes #34)

### LINKED ISSUE #34: Improve the lists chapter
Towards the end of the chapter on lists, particularly on the page `FunctionsAndMethodsForLists`, the quality is not great. A lot of information is dumped on the user and it's not easy for them to process it. There's lots of reading and not much interactivity.

Here is what I'm trying to accomplish:

- Summarise the essential things you can do with lists, like len and append. This is to refresh their memory in the moment and also provide one easy place that they can come back to later if they need a reminder.
- Show that there are many other useful functions available for lists (and also strings). They *don't* actually need to memorise any of these - they just need to know that many useful functions exist, and have a vague sense of the kinds of things they do so that when they encounter a use case they have a chance of guessing that a function might exist and can be found with google.
- Show how to look things up, e.g. with google.

The reason there are no exercises and only a light encouragement to "Try them out in the shell" is because remembering the actual functions is not important. But the way the page is now is lots of dry, boring reading. That's bad.

I'm looking for opinions on how to deal with this. Should we:

- Add some steps where users are required to try out functions in the shell, or even complete exercises? This makes the section less boring. But it also puts more pressure on the user to learn information that is not essential at this stage. I've described my concerns about that [here](https://github.com/alexmojaki/futurecoder/issues/23#issuecomment-668052076). Maybe it's fine, and this is just the time to learn those things.
- Emphasise more and earlier that they don't need to memorise the functions so that they feel comfortable skimming through the list if it's boring.
- Do something completely different? I'm keen for ideas.

Then there's this section:

> Note that in most cases, methods which modify a list in place (`append`, `insert`, `remove`) merely return `None`, while the remaining functions/methods return a new useful value without changing the original argument. The only exception is the `pop` method.
> 
> Modifying a value directly is called *mutation* - types of values which can be mutated are *mutable*, while those that can't are *immutable*. Strings are immutable - they don't have any methods like `append` or even subscript assignment. You simply can't change a string - you can only create new strings and use those instead. That means that this is a useless statement on its own:
> 
>     word.upper()
> 
> The string referred to by `word` isn't modified, instead `word.upper()` returned a new string which was immediately discarded. If you want to change the value that `word` refers to, you have to assign a new value to the variable:
> 
>     word = word.upper()
> 
> Or you can use `word.upper()` immediately in a larger expression, e.g.
> 
>     if word.lower() == 'yes':

This is all very important information that is worth learning properly. Adding some interactive steps (e.g. showing that `word.upper()` on its own doesn't change `word`) would be great. I think the only reason I didn't was that I was tired and lazy. It should maybe be on its own page, the reason I hesitate is that I refer to method names like `pop` and it's useful to be able to scroll up and see the definitions of those methods. They can still use the Previous and Next buttons if the pages are separated though, and that might even be better.

The pages `EqualsVsIs` and `ModifyingWhileIterating` are also heavy reading and not great, but maybe less of a priority.

spamegg1:
I ran into some issues, you can see them in `TODO`s.

### DIFF
--- a/backend/main/chapters/c06_lists.py
+++ b/backend/main/chapters/c06_lists.py
@@ -842,7 +842,7 @@ Run the following code:
             nums.append(4)
             print(nums)
 
-    class pop_remove_index_subscript_assignment(ExerciseStep):
+    class pop_remove_index_subscript_assignment(VerbatimStep):
         """
 As you see, `+` does not modify `nums`, but `append` does. (`new_nums` is not affected by the change in `nums`.)
 
@@ -857,8 +857,7 @@ Now you will solve four exercises about the new functions and methods you learne
 Below is a list of correct and incorrect lines of code, mixed together.
 Each exercise has a solution that includes one (and only one) of the lines below.
 In each exercise you need to find the correct line from the list below that achieves the result described,
-and use it to write the solution. (Sometimes you can use the correct line exactly as it is,
-sometimes you will have to add a little bit more to it.)
+and replace the indicated part with it to write the solution.
 
     x[len(x)] = x[0]
     x[len(x) - 1] = x[0]
@@ -879,58 +878,54 @@ sometimes you will have to add a little bit more to it.)
     x.pop(x.index(0))
     x.index(x.pop(0))
 
-Write a program which takes a list `x` of strings, and modifies `x` to move the first element to the end,
+First exercise:
+
+Write a program which takes a list `x`, and modifies `x` to move the first element to the end,
 keeping the length of `x` the same, then prints `x`.
 For example, if `x = ['a', 'b', 'c']` then the code should modify `x` to `['b', 'c', 'a']` and print it.
+Here is the provided code for you to use. Replace the indicated part with a line from the list above, then run the code.
 
-The solution has exactly three lines of code, one of which is from the list above.
-Find the correct line, write your solution in the editor, and then run the code.
+    __copyable__
+    x = ['a', 'b', 'c']
+    (insert_one_line_from_above)
+    print(x)
         """
 
-        # TODO: this was the only way I could make it to pass. It doesn't like mutating x as part of solution.
-        # I tried writing generate_inputs as in list_contains_exercise but could not make it work.
-        # Probably I should be trying to emulate list_insert(Step) but I'm not sure how to write the check method.
-        # I had to make x a List[str] or List[int] to avoid "solution is missing one positional argument" error.
-        # Same issues continue down in all the exercises, sorry about that.
-        # It would be great if it worked on any kind of list.
-        # Also would be nice to have some MessageStep s for similar but wrong choices.
-        def solution(self, x: List[str]):
-            y = x.copy()
-            y.append(y.pop(0))
-            print(y)
+        program_in_text = False
 
-        tests = [
-            (['This', 'is', 'a', 'list'], ['is', 'a', 'list', 'This']),
-            (['The', 'quick', 'brown', 'fox', 'jumps'], ['quick', 'brown', 'fox', 'jumps', 'The']),
-        ]
+        def program(self):
+            x = ['a', 'b', 'c']
+            x.append(x.pop(0))
+            print(x)
 
-    class subscript_assignment_exercise(ExerciseStep):
+    class subscript_assignment_exercise(VerbatimStep):
         """
 Good job. Next exercise:
 
-Write a program that takes a list `x` of strings and modifies `x` to overwrite
+Write a program that takes a list `x` and modifies `x` to overwrite
 the last element with the first element, keeping the length of `x` the same, then prints `x`.
 For example, if `x = ['a', 'b', 'c']` then the code should modify it to `['a', 'b', 'a']` and print it.
+Here is the provided code for you to use. Replace the indicated part with a line from the list above, then run the code.
 
-The solution is exactly three lines of code, one of which is from the list above.
-Find the correct line, write your solution in the editor, and then run the code.
+    __copyable__
+    x = ['a', 'b', 'c']
+    (insert_one_line_from_above)
+    print(x)
         """
 
-        def solution(self, x: List[str]):
-            y = x.copy()
-            y[len(y) - 1] = y[0]
-            print(y)
+        program_in_text = False
 
-        tests = [
-            (['This', 'is', 'a', 'list'], ['This', 'is', 'a', 'This']),
-            (['The', 'quick', 'brown', 'fox', 'jumps'], ['The', 'quick', 'brown', 'fox', 'The']),
-        ]
+        def program(self):
+            x = ['a', 'b', 'c']
+            x[len(x) - 1] = x[0]
+            print(x)
 
-    class negative_index_concatenation_exercise(ExerciseStep):
+    class negative_index_concatenation_exercise(VerbatimStep):
         """
 Excellent!
 
-You might realize that accessing the last element via `x[len(x) - 1]` is a bit cumbersome. The same can be achieved by `x[-1]`.
+You might realize that accessing the last element via `x[len(x) - 1]` is a bit cumbersome.
+The same can be achieved by `x[-1]`.
 Similarly, the second to last element `x[len(x) - 2]` is equivalent to `x[-2]`, and so on.
 Python allows us to count the index backwards too, starting at the last element with `-1`:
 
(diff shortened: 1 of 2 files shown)


backend/main/chapters/c06_lists.py
```diff
@@ -829,16 +826,367 @@ class sum_list(Step):
 - **`subscripting`**: Get a value at an index. `nums[0]` is 1, `nums[1]` is 2, `nums[2]` is 3.
 - **`+`**: Concatenates lists. `nums + [4, 5]` is `[1, 2, 3, 4, 5]`.
 
-Here's some new things. Try them out in the shell.
+Note that `nums.append(4)` modifies the existing list `nums`, while `nums + [4, 5]` does not.
+To preserve the value of `nums + [4, 5]` we must assign it to a *new variable*.
+Type and run the following code in the editor:
+
+__program_indented__
```
alexmojaki, line 833:
```suggestion
Run the following code:

    __copyable__
    __program_indented__
```

Typing in this code doesn't provide much learning benefit. It's tedious and all the syntax is familiar.

backend/main/chapters/c06_lists.py
```diff
@@ -829,16 +826,367 @@ class sum_list(Step):
...
+            print(new_nums)
+            print(nums)
+            nums.append(4)
+            print(nums)
+
+    class pop_remove_index_subscript_assignment(ExerciseStep):
+        """
+As you see, `+` does not modify `nums`, but `append` does. (`new_nums` is not affected by the change in `nums`.)
+
+Here's some new things. Try them out in the shell. Again suppose we have a list `nums = [1, 2, 3]`.
```
alexmojaki, line 848: Once we merge https://github.com/alexmojaki/futurecoder/pull/73 it might be good to make actual steps of these with predicted outputs. In fact if we'd done that first we might not have done all this. But the combination is also good.

backend/main/chapters/c06_lists.py
spamegg1: OK.

backend/main/chapters/c06_lists.py
spamegg1: Hmm, #73 is a complicated idea. I'm not sure what I think about it. It might be good if it's used sparingly.

backend/main/chapters/c06_lists.py
```diff
@@ -829,16 +826,367 @@ class sum_list(Step):
 - **`subscripting`**: Get a value at an index. `nums[0]` is 1, `nums[1]` is 2, `nums[2]` is 3.
 - **`+`**: Concatenates lists. `nums + [4, 5]` is `[1, 2, 3, 4, 5]`.
 
-Here's some new things. Try them out in the shell.
+Note that `nums.append(4)` modifies the existing list `nums`, while `nums + [4, 5]` does not.
+To preserve the value of `nums + [4, 5]` we must assign it to a *new variable*.
```
alexmojaki, line 830: Not strictly true. It could be assigned to the same variable, or just used in some other way.

alexmojaki:
OK, this is making me question some things and I'm not sure what the answers are.

I initially said these should be ExerciseSteps thinking it was OK if they creatively came up with a different one-line solution. Now I'm thinking this exercise is more about understanding existing code than creative problem solving. Therefore they should be forced to use an existing line exactly, and we use VerbatimSteps. What do you think?

(I think I can solve the problem you came across about modifying inputs quite easily, that alone should not be a reason to avoid ExerciseSteps)

Either way I think it's better if we simply describe each exercise something like:

> Complete this program:

```python
x = ['a', 'b', 'c']
<insert one line from above>
print(x)
```

> The program should <do thing to x> and therefore print <bca or whatever>.

I'm very unsure about how to present the `x + [x[0]]` problem.

1. I'm nervous about your "Sometimes you can use the correct line exactly as it is, sometimes you will have to add a little bit more to it". I think it takes the exercises too far beyond 'fun and easy'. It's easy to guess that `x + [x[0]]` is incomplete but the student now has to wonder if any of the other solutions are incomplete and in what way.
2. We could present the solutions in the list as `x = x + ...` etc. But then students might look at this and wonder why it's not written `x += ...`. They might even discover through experimentation that `+=` actually modifies the list (and is different from `x = x + ...`) which could lead to all sorts of confusion that we're not ready to explain. Maybe I'm being paranoid.
3. We could leave the solution as is but present the template to fill in as `x = <insert line>`. Still creates the problem above.
4. Make the template `y = <insert line>; print(y)`. My initial concern was that it's too obvious just looking at this that the solution involves `+` after we just emphasised how you should put the result of `+` into a new variable. But as I write this I think that's fine, it means they paid attention and learned something, and exercises are meant to link to learning material. What do you think?

backend/main/chapters/c06_lists.py
spamegg1: OK, I'll say `we *can* assign...` or `one of the ways to preserve the value is to assign it...`

spamegg1:
In my initial markdown draft, my intention was for them to find the correct line and run just that line in the shell. (It could even be a multiple choice quiz.) Then I followed your 3-line solution idea so this is the result.

1. I agree, it makes them second guess the simple solution they might have come up with. But I could not find another way to make the instructions precise. Without explanation, they may also be under the impression that they need to use the line exactly. Or they may not quite understand the difference between, say `x + [x[0]]` and `y = x + [x[0]]` yet.

2. In my latest draft I actually changed them to `x = ...` first, but then ran into the issue with the `ExerciseStep`s so I changed them back and decided to hear your take on it first. Also `x = x + [x[0]]` *does* modify the original list so they could just do the `append` solution instead. To nudge them away from `append` I had to say *don't modify, create new list*.

3. Agreed.

4. I think it's OK too. These are supposed to be short quick exercises anyway.

alexmojaki:
> In my initial markdown draft, my intention was for them to find the correct line and run just that line in the shell.

Yes, but that can get weird, e.g. if they don't define `x` they'll get an error. And if they modify the list they have to check it after. It's similar to https://github.com/alexmojaki/futurecoder/issues/31#issuecomment-670810030

> (It could even be a multiple choice quiz.)

That might be nice but:

1. Typing may help the learning process
2. If they get the answer wrong, they can see what's wrong in the output
3. This hasn't been implemented in the text

> I would lean towards a solution that is close to: "choose the correct line" and not much more work than that.

One thing the can help is to make the provided template copyable. Also it should be valid syntax so that it gets highlighted, e.g.

```python
__copyable__
x = ['a', 'b', 'c']
(insert_one_line_from_above)
print(x)
```

Anyway, this doesn't resolve my main question:

> I initially said these should be ExerciseSteps thinking it was OK if they creatively came up with a different one-line solution. Now I'm thinking this exercise is more about understanding existing code than creative problem solving. Therefore they should be forced to use an existing line exactly, and we use VerbatimSteps. What do you think?

> Also x = x + [x[0]] does modify the original list

It doesn't, it replaces the variables with a new list and the original is lost. Try:

```python
x = [1, 2, 3]
y = x
x = x + [4]
print(x)  # [1, 2, 3, 4]
print(y)  # [1, 2, 3]
```

But change it to `x += [4]` and `y` will change as well, like `append`.

spamegg1:
I'm OK with forcing them to use an existing line exactly. It didn't occur to me to use `VerbatimStep`s, otherwise I would have done that. 

Would you like me to go ahead and make a commit where I change them all to `VerbatimStep` with copyable blocks where they have to replace one line?

Interesting, I didn't know that `x +=` and `x = x +` were different. I thought the first was a short-hand for the second and that they both mutated `x`. So should I simply add `x = ` to those statements?

Give me the go-ahead and I'll do it.

alexmojaki:
> I'm OK with forcing them to use an existing line exactly. It didn't occur to me to use `VerbatimStep`s, otherwise I would have done that. All my issues came from struggling with `ExerciseStep`s anyway.
> 
> Would you like me to go ahead and make a commit where I change them all to `VerbatimStep` with copyable blocks where they have to replace one line?

Yes.

> Interesting, I didn't know that `x +=` and `x = x +` were different. I thought the first was a short-hand for the second and that they both mutated `x`.

Yes, it's a nasty corner. `+=` actually first tries to call `x.__iadd__`. Hopefully students won't notice this because at this stage it'll cause extra confusion.

> So should I simply add `x = ` to those statements?

Leave the listed solutions as they are, make the template contain `y = ...` as in point 4 above.

alexmojaki:
OK, I think this is more wordy and repetitive than it needs to be. I've suggested my wording in https://github.com/spamegg1/futurecoder/pull/1. If you think that's OK, please try and make similar changes in the second page of exercises.

Some messagesteps I'm considering:

- Check if they wrote `x = x + [x.pop(0)]` or `x += [x.pop(0)]` instead of using append.
- Check if they used `x[-1]` immediately
- Check if they write more than three statements
- Check if they leave out `y = ` in the last exercise

All of these are in case they don't follow instructions properly, probably not very important.

I do think it's worth having hints in these exercises. The first hints should make sure they're following the instructions correctly.

spamegg1:
Haha, usually I am very terse and you end up adding a lot of words... now I've gone too far in the other direction! Merged it.

Will add back my hints, better wording on second page, and attempt some `MessageStep`s tomorrow. Gotta sleep now. It'd be nice if we can finish this soon, I'd like to disappear for a week or so.

alexmojaki:
Great!

I've enabled tables and added them in https://github.com/spamegg1/futurecoder/pull/2

spamegg1:
Let me see if I understand things clearly.

If I want to catch a correct but unintended solution with a `MessageStep`, the outer level `check` must return `True`, and I need to use `after_success = True`, and the `check` method of the `MessageStep` should also return `True` in order to display the message.

But a `VerbatimStep` requires the AST to match exactly. So the outer level `check` would never return `True` in the case of a correct but different solution.

Therefore... in order to use a `MessageStep` inside a `VerbatimStep` I have to convert the `VerbatimStep` to a `Step` and manually write a `check` method for it, so that it can return `True` without matching the AST exactly but still being a correct solution.

So... I would have to undo our decision of going with `VerbatimStep`.

Am I on the right track so far?

I was trying some things to catch the use of concatenation in the first exercise. Would something like this work? Also how do I explain it to the user? It seems pretty complicated for beginners (even I was befuddled yesterday).

```python
class used_concatenation(MessageStep):
    """
You wrote `x = x + [x.pop(0)]` (or `x += [x.pop(0)]`).
This achieves the same result on the surface, but it does not modify the original list as intended.
Remember that `+` creates a new list. Then this new list gets assigned to the variable `x`,
and the original list is lost.
    """
    after_success = True

    def program(self):
        x = ['a', 'b', 'c']
        x = x + [x.pop(0)]
        print(x)

    def check(self):
        return search_ast(
            self.stmt,
            ast.Call(func=ast.Name(id='+')),
        )
```

alexmojaki:
Your reasoning is all correct except for one thing. Just don't set `after_success = True`, and then both steps can be verbatim. `after_success = True` is purely a convenient way to say "the outer check succeeded" which is common and would otherwise require calling `super` or something. It doesn't have to mean 'the user did the right thing' in som special sense. Besides, they didn't follow the instructions because they didn't use a line from the list exactly, so they didn't succeed.

alexmojaki:
Your explanation is perfect, but it only applies to `+=`. `+=` requires a separate message along the lines of "this does the right thing, but you have to choose a line from the list as instructed".

spamegg1:
I see. It's imperative to stick with `VerbatimStep` to enforce they use exactly a line from the list.

OK... so `VerbatimStep` inside another `VerbatimStep`? I'll try that.

alexmojaki:
The hints should all start by reiterating the instructions, something like:

> To be clear, your solution should have exactly three statements: `x = ...`, one line copied exactly from the list (no additions), and `print(x)`.

Adjust for the `y = ...` exercises.

In general, once all hints have been revealed, there should still be some work/thinking left for the user. The hints shouldn't basically give away the answer, because that's what the Show Solution button is for at the end.

In this particular case that means the hints should reveal very little information. Keep them vague and only give a few. I don't think the function name should be revealed, that typically narrows it down to about 4 options.

alexmojaki:
Nice work!

PR #77, opened 2020-10-10
Fancy tracebacks

### LINKED ISSUE #39: Fill the right half of the screen with the shell until the editor is introduced
Instead of having it only in the bottom right.

Originally mentioned in #31

### LINKED ISSUE #75: Better tracebacks
Currently I use https://github.com/alexmojaki/stack_data to format tracebacks, and they look pretty good:

traceback

We can definitely do more. I'd like to make these as good as possible. Some ideas:

- Explain the exception type and error message in more detail for beginners. https://github.com/aroberge/friendly-traceback is a good option for this. I don't want to use the exact format of friendly-traceback, so I'd like to just extract the explanation and any other useful bits. @aroberge has indicated in https://github.com/aroberge/friendly-traceback/issues/96#issuecomment-698004351 that he is interested in working on this but may not have time for quite a while, and I'd rather not wait around, at least not without trying first to see what can be done easily.
- Show variables and other simple expression values in each frame. stack_data already provides the machinery for this.
- Show suggestions for fixes using https://github.com/SylvainDe/DidYouMean-Python
- Suggest fixes for syntax errors:
    - https://github.com/cfbolz/syntaxerrors
    - https://github.com/naturalness/sensibility - see https://github.com/alexmojaki/futurecoder/issues/70
- Get the exact node executed where possible (already provided by stack_data) and use it to add even more info like in https://github.com/aroberge/friendly-traceback/issues/96#issuecomment-697999929
- Make it pretty ~with https://github.com/willmcgugan/rich~ in the frontend 

Personally I would rather develop and test functionality here first and worry about integrating it back into the appropriate library like friendly-traceback or DidYouMean later. It's more difficult to develop a polished generic API than one concrete use case. But @aroberge seems to prefer the opposite.

alexmojaki:
This closes #75 and overhauls tracebacks to be much more helpful, beginner friendly, and nice to look at.

Here is an example script which demonstrates most of the functionality:

```python
def print_many(n, thing):
    for _ in range(n):
        print(thing)
        try:
            foo()
        except:
            thong

def foo():
    return foo()

def print_twice(x):
    print_many(
        2,
        x,
    )


print_twice("Hello")
```

Main features:

- Values of variables in a table in each frame
- Explanations of errors using https://github.com/aroberge/friendly-traceback - hover over the 'i' icon next to an exception message to see.
- Suggestions for fixes from https://github.com/SylvainDe/DidYouMean-Python
- Highlighting of the executing node in blue (although that was already there before)
- Handling of exception chains, e.g. "During handling of the above exception, another exception occurred:"
- Handling of chains of similar frames in recursion errors, i.e. "Similar frames skipped:"

Lots of small decisions made in the design:

- Start with "Error traceback:" rather than "Traceback (most recent call last):". The former makes it clear that this is an error while still introducing people to the jargon 'traceback'. The latter is less clear to beginners. The information 'most recent call last' is more noisy than helpful.
- Each frame is outlined because that matches the word 'frame' nicely and implies an isolated scope.
- The line numbers are styled just like the editor above to indicate that users are looking at a sample of their code.
- Frames not belonging to the user's program are hidden. Showing code from e.g. the standard library would not be very useful to a beginner.
- The header of each frame contains only the name of the current function (or other code object). There's no need for the filename since only the user's 'file' is shown, and the line number is already on the side.
- The header is omitted when it would be `<module>` as this is not meaningful to beginners and doesn't really add much information anyway.
- Extra lines of code for context are not added as in some other tracebacks (e.g. python) since users just wrote this exact code and can see it right above so it doesn't add much value. However stack_data ensures that multiline expressions are shown in full.
- A bit of red text makes it clear this is an error, but it's not all in red to not scare the user.
- Exception message is in slightly larger text for emphasis, with the type in bold, since this is what the user needs to look at the most.
- An exception in the shell and not in a deeper function call only shows the exception message (and maybe suggestions). Showing the frame would be pointless.
- Syntax errors are treated differently. The explanation from friendly traceback is shown immediately (rather than in an icon that needs hovering) because the user probably needs to look at it more so than a runtime error, and there's not much else to see anyway.

Also threw in some other random little fixes not worth their own PR. In particular this closes #39

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3.8-distutils \
     gcc \
     postgresql-client-12 \
+    git \
     libpq-dev
 
 RUN ln -s /usr/bin/python3.8 /usr/local/bin/python
--- a/backend/main/chapters/c01_the_shell.py
+++ b/backend/main/chapters/c01_the_shell.py
@@ -1,5 +1,8 @@
 # flake8: NOQA E501
-from main.text import MessageStep, Page, Step, VerbatimStep
+import ast
+from textwrap import dedent
+
+from main.text import MessageStep, Page, Step, VerbatimStep, search_ast
 
 
 class IntroducingTheShell(Page):
@@ -37,20 +40,20 @@ Try doing some more calculations now. You can multiply numbers with `*`, divide 
 
         program = "5 - 6"
 
-        class x_to_multiple(MessageStep):
-            """
+        def check(self):
+            try:
+                return search_ast(self.stmt, (ast.Mult, ast.Div, ast.Sub))
+            except SyntaxError:
+                if "x" in self.input:
+                    return dict(
+                        message=dedent(
+                            """
             I see an 'x'. If you're trying to multiply, use an asterisk, e.g:
 
                 3 * 4
             """
-
-            program = "3 x 4"
-
-            def check(self):
-                return "x" in self.input
-
-        def check(self):
-            return self.input_matches(r"\d[-*/]\d")
+                        )
+                    )
 
     final_text = """
 Excellent! Keep experimenting. When you're ready, click 'Next' to continue.
--- a/backend/main/chapters/c01_the_shell.py
+++ b/backend/main/chapters/c01_the_shell.py
@@ -5,7 +5,7 @@ from main.text import MessageStep, Page, Step, VerbatimStep
 class IntroducingTheShell(Page):
     class first_expression(VerbatimStep):
         """
-At the bottom right of the screen is the *shell*. This is a place for running small bits of Python code. Just type in some code, press enter, and it'll run! Try it now:
+On the right is the *shell*. This is a place for running small bits of Python code. Just type in some code, press enter, and it'll run! Try it now:
 
 1. Click anywhere on the shell (the black area).
 2. Type `__program__`
--- a/backend/main/chapters/c07_nested_loops.py
+++ b/backend/main/chapters/c07_nested_loops.py
@@ -124,6 +124,19 @@ Make sure each line is in the correct loop and has the right amount of indentati
 
         parsons_solution = True
 
+        def check(self):
+            try:
+                return super().check()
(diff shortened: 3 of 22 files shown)


alexmojaki:
@spamegg1 check it out!

spamegg1:
Very nice! 

(This reminds me of Harvard's CS50's attempt at making `gcc` errors and `man-pages` more beginner-friendly.)

I could not find any obvious issues. A few examples:
```
>>> word
'Hello'
>>> wird
NameError: name 'wird' is not defined
Did you mean...

    'word' (local)?
```
Hovering over the `i` shows:
```
A NameError exception indicates that a variable or function name is not known to Python. Most often, this is because there is a spelling mistake. However, sometimes it is because the name is used before being defined or given a value.

In your program, the unknown name is wird.
```
Another:
```
>>> 'Hello " + your_name
    'Hello " + your_name
                       ^
SyntaxError: EOL while scanning string literal
at line 1

A `SyntaxError` occurs when Python cannot understand your code.

You starting writing a string with a single or double quote
but never ended the string with another quote on that line.
```
Another:
```
>>> print word + ' ' + your_name
    print word + ' ' + your_name
          ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(word + ' ' + your_name)?
at line 1

A `SyntaxError` occurs when Python cannot understand your code.

Perhaps you need to type

     print(word + ' ' + your_name)

In older version of Python, `print` was a keyword.
Now, `print` is a function; you need to use parentheses to call it.
```
Another:
```
Error traceback:

2 | x + x.pop(0)
-- | --

x | ['b', 'c']
-- | --
TypeError: can only concatenate list (not "str") to list
```
Hovering over the `i` says:
```
A TypeError is usually caused by trying to combine two incompatible types of objects, by calling a function with the wrong type of object, or by tring to do an operation not allowed on a given type of object.

You tried to concatenate (add) two different types of objects: a list and a string (str)
```

ChrisjanWust:
Fuck me this is cool!

alexmojaki:
Thanks reviewers!

PR #88, opened 2020-10-20
Pyflakes additions

josephmarcus9:
linting module and testing script


backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
     UnusedVariable: """
 **Unused variable `{0}`**
-
 You defined a variable `{0}` but never used it. Did you forget to use it?
 Maybe you used the wrong variable in its place? If you don't need it, just remove it entirely.
-    """
+    """,
+    IsLiteral: """
+**Is literal**
+You used the is/is not statement for comparison. You should have rather used the == / != statements,
```
alexmojaki, line 19:
```suggestion
You used the is/is not statement for comparison. You should have rather used the `==` / `!=` statements,
```

This is markdown, make code look good.

backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
 
 MESSAGES = {
+    UnusedImport: """
+**Unused import `{0}`**
+You imported a module `{0}` but never used it. Did you forget to use it?
+Maybe you used the wrong import in its place? If you don't need it, just remove it entirely.
+    """,
     UnusedVariable: """
 **Unused variable `{0}`**
-
```
alexmojaki, line 9: Keep the blank lines so that the title shows on a different line from the body. A blank line separates paragraphs in markdown.

backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
+require instead.   
+""",
+    ImportStarNotPermitted: """
+
+**Import made using * **
+This * import is used to import everything from a designated module under the current 
+module, allowing the use of various objects from the imported module- without having to prefix them with the module's 
+name. Refrain from using this type of import statement and rather explicitly import a few statements that you may 
+require instead.
+""",
```
alexmojaki, line 58: Try to emit the same warning only once. If it's not possible, explain why and use a variable instead of copying the whole message.

backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
 }
 
 
-def lint(tree):
+def lint(code):
     # Wrap the whole module in a function
     # so that pyflakes thinks global variables are local variables
     # and reports when they are unused
-    function_tree = ast.parse("def f(): 0")
-    function_tree.body[0].body = tree.body
```
alexmojaki, line 21: The comment explains why this is needed, don't remove it.

backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
-
+    function_tree = ast.parse(code)
     ch = checker.Checker(function_tree, builtins=["assert_equal"])
     ch.messages.sort(key=lambda m: m.lineno)
     for message in ch.messages:
-        if type(message) in MESSAGES:
-            message_format = MESSAGES[type(message)]
+        if type(message) in MESSAGES:  # type(message) is the key component of dictionary
+            message_format = MESSAGES[type(message)]  # message_format is the value component of the dictionary
```
alexmojaki, line 86: These comments are redundant.

backend/main/linting.py
```diff
@@ -1,28 +1,89 @@
...
 **Unused variable `{0}`**
-
 You defined a variable `{0}` but never used it. Did you forget to use it?
 Maybe you used the wrong variable in its place? If you don't need it, just remove it entirely.
-    """
+    """,
+    IsLiteral: """
+**Is literal**
+You used the is/is not statement for comparison. You should have rather used the == / != statements,
+which are used to compare constant literals (str, bytes, int, float, tuple)- referring to equality.
```
alexmojaki, line 20: `==` can be used to compare anything, it's just a bad idea to use `is` for literals.

backend/main/linting_tester.py
```diff
@@ -0,0 +1,69 @@
...
+    pass
+"""
+
+    code_multi_value_repeated_key = """
+my_dict={'a':1,'b':4,'b':2}
+
+"""
+
+    for x in lint(code_unused_var):
+        print(x)
```
alexmojaki, line 69: Convert this to a proper unit test under tests.py so that developers don't need to edit the code and inspect the output to know if it works.

alexmojaki (changes requested):

backend/main/linting.py
josephmarcus9: Cool, understand

backend/main/linting.py
josephmarcus9: Ok changed all of them to this format!

backend/main/linting.py
josephmarcus9: Can You please elaborate on what you mean by emitting the same warning once. Does that have to do with the fact that some commands may result in 2 warnings being thrown?

backend/main/linting.py
josephmarcus9: Ok will remove, had them initially for my understanding

backend/main/linting.py
alexmojaki: `for x in lint(code_import_star):` prints the same message twice. See if we can have fewer entries in this dict and still get a warning whenever they import star.

backend/main/linting.py
josephmarcus9: Ok, should I restore the `lint()` function back to how it was?

backend/main/linting.py
josephmarcus9: Ok added some more to the explanation

backend/main/linting.py
alexmojaki: Yes

backend/main/linting.py
alexmojaki: Yes

backend/main/linting.py
josephmarcus9: Ok taking out the `ImportStarUsage `entry. The `ImportStarUsed` entry should cover all instances of a star being imported.

backend/main/linting.py
josephmarcus9: Ok cool, will do. Had to change it for testing purposes. Out of interest, when restored to its original state how could I test the function -lint(tree)

alexmojaki: By passing it a tree. Just make a utility function in your tests.

Issue #61, opened 2020-09-06
Chapter about quotes and f-strings


alexmojaki:
I'd like to add a chapter, tentatively called "A bit more about strings", after the chapter on lists and before nested loops, with two pages:

1. Double quotes. So far I've only used single quotes, leaving students probably wondering how to write an apostrophe inside a string. They will learn about escape sequences later. This page should mention that both single and double quotes are allowed and they do the same thing. `"abc" == 'abc'` (they should actually see this), they both get reduced to single quotes in the shell, and become the same thing when printed. Double quotes are useful when you want your string to contain single quotes and vice versa. All pages prior to this page should only use single quotes, which I think currently isn't the case.
2. f-strings. In the nested loops chapter I introduce `print` with multiple arguments which was a bad idea. Instead students should be shown the basics of f-strings and encouraged to use them instead of `+` in most situations. In particular they should be shown that adding a string and a number doesn't work but f-strings get around this. Then f-strings should be used in following pages where sensible, including modifying existing pages.

@BobAnkh you seemed interested in writing course content. This should be fairly simple, want to give it a go?

BobAnkh:
Ok, I will try to do it in these days.

alexmojaki:
Great, thanks! See https://github.com/alexmojaki/futurecoder/blob/master/how_to_contribute.md#how-to-implement-course-content-in-python for some guidance on how writing a chapter works.

BobAnkh:
Ok, thank you for your guidance. I will see it.

Kush1101:
@alexmojaki If the issue is open for contribution, I'd like to give it a try.

BobAnkh:
@alexmojaki I'm so sorry that I'm busy this week. So if @Kush1101 want to do it, I think it is better to let him take this.

alexmojaki:
Hi @Kush1101, nice to meet you! Thanks for being willing to contribute!

@BobAnkh don't feel bad, I am grateful that you want to help. This issue is not urgent.

Ideally I would like everyone to have something they can work on (when they have time) to maximise productivity. It would be a pity if @Kush1101 took this issue and then when @BobAnkh had time there was nothing for him to do.

Here are some other topics I'd like to cover soon:

- Nested lists, under the nested loops chapter
- `input()`
- Types (see the points in #23)

Are either of you interested in trying to write something about any of these topics? If not then that's fine too, it's just a suggestion. No one should say they will attempt something they don't want to do.

[...]

PR #96, opened 2020-10-24
Chapter about quotes and f-strings #61

spamegg1:
- added `c07_a_bit_more_about_strings.py`
- renumbered `nested_loops`, `functions` and `boolean_operators` chapters
- tweaked some exercises in `c06_lists.py`


spamegg1:
Some issues remain:

How to use `__copyable__` with plain text that doesn't follow Python syntax?
```
    __copyable__
    Special cases aren't special enough to break the rules.
```
gives
```
File "/usr/src/app/backend/main/utils/__init__.py", line 168, in run
    node.set("class", node.get("class") + " copyable")
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

- Similar issue with the broken program exercise at the very end of f-strings page.
- Could not figure out how to include broken string `'Alice's Diner'` and make the build pass. Does it need a `Step` with a manual `check` method?

alexmojaki:
> How to use `__copyable__` with plain text that doesn't follow Python syntax?

Looks like a bug. Try fixing it!

>  Does it need a Step with a manual check method?

Yes, a verbatim step requires valid syntax.

spamegg1:
OK, I was able to fix the `'Alice's Diner'` issue. I hope `return "SyntaxError" in self.result` is good enough of a check! 

However now `__program_indented__` does not get syntax highlighting. I assume this is normal behavior, as in Chapter 4:
```
for character in name:
print(character)
```

Still working on the `__copyable__` issue. I assume when you say "bug" you mean in my code, not in the API. According to the error `node.get("class")` is returning `None` when it shouldn't??? Still trying to understand what nodes are.

alexmojaki:
> I hope return "SyntaxError" in self.result is good enough of a check!

It's a start, but try making it more robust.

> However now __program_indented__ does not get syntax highlighting. I assume this is normal behavior, as in Chapter 4:

Invalid syntax doesn't get highlighted, that's actually related to the copyable bug.

> Still working on the `__copyable__` issue. I assume when you say "bug" you mean in my code, not in the API. According to the error node.get("class") is returning None when it shouldn't??? Still trying to understand what nodes are.

No it's a bug in my code. Node means HTML node. It doesn't have a class so it returns None. Try `node.get("class", "")` or `node.get("class") or ""`.

spamegg1:
I've tried `node.get("class", "") or ""` but now I get:
```
File "/usr/src/app/backend/main/text.py", line 249, in __init__
    assert "__copyable__" not in str(cls.step_texts)
AssertionError
```

If I comment out the `assert` statement obviously the build passes but it breaks things (`__copyable__` is displayed on top of the text, also gets copied)

alexmojaki:
Try actually learning how the code works. See what happens when the syntax is valid. This is a great opportunity to learn and to contribute to the actual functionality in addition to course content.

spamegg1:
Yeah that's true, it's really hard though.

alexmojaki:
OK, I'm done. Let me know if you're fine with the changes. You can see my changes by clicking on the specific commits.

alexmojaki:
Also please add those message steps.

spamegg1:
Changes all look good. Your text makes things clearer, also gives fewer hints for the final exercise.

Do you think "replacing the concatenation (`+`) with an f-string" might be a bit confusing? There are multiple `+`s and they might think they have to replace each one with an `f`. Maybe I'm overthinking this.

We decided to skip the message steps, anything else I should do?

alexmojaki:
I think you're right, how about

> Fix this by replacing all the string concatenations (+) with a single f-string

spamegg1:
OK let me commit that.

alexmojaki:
Awesome! You've now done two full chapters!

Now to replace all prints with multiple arguments with f-strings.

spamegg1:
> Now to replace all prints with multiple arguments with f-strings.

Right!

alexmojaki:
In the nested loops chapter, the first three steps can be reduced to two or maybe even just one.

spamegg1:
OK, I'll open an issue first with markdown drafts.

PR #98, opened 2020-10-31
Added f-strings to Chapters 8 and 9

spamegg1:
***Data migration warning!*** Deleted 2 steps from Chapter 8, first page "Introducing Nested Loops"
- also did a few `TODO message` steps

### DIFF
--- a/backend/main/chapters/c08_nested_loops.py
+++ b/backend/main/chapters/c08_nested_loops.py
@@ -12,52 +12,95 @@ class IntroducingNestedLoops(Page):
 You've seen that the indented body of an `if` or a loop can contain any kind of statement, including more `if` statements and loops. In particular a loop can contain another loop. Here's an example:
 
 __program_indented__
-        """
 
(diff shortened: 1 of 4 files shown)


alexmojaki:
Excellent work! Your message step was fine, you just found a bug in the API. Are you happy to merge?

spamegg1:
Yes, happy to merge!
I was actually going to tell you... I suspected a bug because... I changed the `check` method to:
```python
def check(self):
    return True
```
and it still would not trigger the message step! Sorry I should have told you earlier.

alexmojaki:
> I know ExerciseStep implements check based on solution but I did not know if MessageStep overrode it.

It doesn't. You can look at the source to check.

alexmojaki:
Checked that no one was on the deleted steps.

spamegg1:
What next?

alexmojaki:
Think about how to explain nested lists: accessing and setting indices, and looping over them. I will create an issue with some of my thoughts.

Issue #100, opened 2020-10-31
Nested lists pages


alexmojaki:
In the nested loops chapter, we need to teach nested lists. This is essential for #3.

Important concepts to cover:

1. Working with nested elements by index:
    1. Getting, e.g. `x = arr[i][j]`.
    2. Setting, e.g.  `arr[i][j] = x`.
2. Looping over all nested elements, e.g. `for sub in arr: for x in sub: print(x)`.

These could be in the opposite order: it might make sense to teach 2 before 1. I haven't figured out all the details.

<details><summary> It's possible to teach the concept of getting by nested indices without mentioned nested lists, because a list of strings is already a nested sequence. Click to see how that could look. Does this make it easier to understand nested lists when they're introduced? Maybe, maybe not.</summary>

Exercise: given a list of strings, print the first letter of the second string in the list. For example, given:

    __copyable__
    strings = ["abc", "def", "ghi"]

you should print `d`.

Solution:

    string = strings[1]
    print(string[0])

---

You may have solved it like this:

    string = strings[1]
    print(string[0])

There's a shorter way. `strings[1]` is an expression like any other, and subscripting like `[0]` can be used on
any expression, not just variables. So you can skip the intermediate variable and just do it in one line:

    print(strings[1][0])

Take a good look at this syntax. If it looks new and fancy, it's not. It's just the usual syntax for subscripting,
applied twice.
</details>

Looping over a nested list can be introduced with a simple example like:

```python
    matrix = [
        [1, 2],
        [3, 4],
    ]

    for row in matrix:
        for num in row:
            print(num)
        print('---')
```

<details><summary>Here is one idea for exercising with loops. The user must search for a substring in a nested list of strings (maybe each list represents a book, and each string is a page) but there are different things that could be done when the user finds the substring. Solving all these exercises would be a lightweight way to train thinking about indenting and setting variables correctly. Click for details. </summary>

First exercise: print every string that contains the word:

    word = "hello"
    strings = [
        [
            "hello there",
            "how are you",
        ],
        [
            "goodbye world",
            "hello world",
        ]
    ]

    for sublist in strings:
        for string in sublist:
            if word in string:
                print(string)

Output:

    hello there
    hello world

Next: print a boolean for each sublist indicating if the word exists:

    for sublist in strings:
        present = False
        for string in sublist:
            if word in string:
                present = True
        print(present)

Finally: just print one boolean indicating if the word is present at all:

    present = False
    for sublist in strings:
[...]

PR #105, opened 2020-11-24
Nested lists pages #100

spamegg1:
#100

### DIFF
--- a/backend/main/chapters/c08_nested_loops.py
+++ b/backend/main/chapters/c08_nested_loops.py
@@ -3,7 +3,14 @@ import ast
 from typing import List
 
 from main.exercises import generate_string
-from main.text import ExerciseStep, MessageStep, Page, VerbatimStep, Disallowed
+from main.text import (
+    ExerciseStep,
+    MessageStep,
+    Page,
+    VerbatimStep,
+    Disallowed,
+    search_ast,
+)
 
 
 class IntroducingNestedLoops(Page):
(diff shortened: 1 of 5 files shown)


spamegg1:
How about this? If not we can revert the commit.

spamegg1:
OK, so you went with removing `book` and changing the lengths of sublists in `numbers` to compensate (the empty sublist is also a nice touch there.). That is fine with me. I wasn't a fan of the `book` idea to begin with. It is shorter and flows better now. Is there anything else you want? I'm happy to merge.

By the way why is this needed? `predicted_output_choices = None` To clear the shell from the previous step?

alexmojaki:
It's because that step inherits from the previous step.

I need to review the last bit of the chapter before merging.

How do you feel about the chapter as a whole?

What made you decide to remove the bit about `strings[2].append(x)`? I'm undecided about it.

spamegg1:
Ah I get it now... When I moved `book` to the second page, it didn't occur to me to simply change that text to `strings[2].append(`... I was fixated on moving that mention from the final text of the first page to somewhere in the second page (where `book` was moved). Seeing I couldn't do that, I removed it! 

It's good as a whole, I'll add that back into the final text again.

alexmojaki:
Ready to merge when you are.

Next task will be to update the `diagonal_winner` exercise to use a nested list.

spamegg1:
:+1:

PR #106, opened 2020-11-29
Rewrote chp10 `diagonal_winner` exercise to use nested lists

spamegg1:
- should we add new hints? I don't think so.

### DIFF
--- a/backend/main/chapters/c10_boolean operators.py
+++ b/backend/main/chapters/c10_boolean operators.py
@@ -560,42 +560,58 @@ and play a few games [here](https://gametable.org/games/tic-tac-toe/).
 We need to check if someone has won a game. Our function `all_equal` is already helpful for checking rows.
 
 Write a function to check if someone has won a game by placing 3 of the same pieces on one of the diagonal lines.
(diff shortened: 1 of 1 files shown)


spamegg1:
Oops... It looks like I forgot to change `def diagonal_winner(row1, row2, row3):` to `def diagonal_winner(board):` in the final text... I shouldn't make a PR for it, could you fix it? Thank you!

alexmojaki:
I notice that you create a new fork each time you make a PR, and delete the previous fork.

You don't need to do that every time. Rather you should be configuring your local repository to be able to pull the latest changes from my side. See https://gist.github.com/Chaser324/ce0505fbed06b947d962 or https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork and https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork

Basically:

```
git remote add upstream https://github.com/alexmojaki/futurecoder.git
git pull upstream master
```

Issue #92, opened 2020-10-21
Next course chapters


alexmojaki:
EDIT: replaced by #165 

This issue is replacing #23 which contains quite a bit of discussion that newcomers don't need.

Here is an overview of what needs to be written next:

- ~Nested lists, under the nested loops chapter~
- ~Basics of types:~
    - ~Different values have different types~
    - ~These can be revealed with the `type()` function~
    - ~Types include `str`, `int`, `float`, `bool`, and `list`.~
    - ~Different types behave differently, e.g. `+` and `<` means different things for each, and each have different methods.~
    - ~Different types are often 'incompatible', e.g. numbers and strings~
- `input()`. Explain that it always returns a string, and must convert to int when needed.
- The tic-tac-toe project described in #3, but without using while loops. This means the code should assume valid user input and the user can only play once per program run.
- `while`. Start with a simple counter (contrast with `for i in range`), then another example with a proper condition (any ideas for a compelling example?), then `while True`. Show how `while` can be used to improve the tic-tac-toe program.

spamegg1:
Types draft. I wrote it with the intention of having it at the end of Chp3 Variables. If that's not suitable I can simply change some of the text.

<details>
<summary>Click to expand</summary>

## Types

### Page: (should it go at the end of the Variables chapter?)

#### Step: different values have different types, `str`

So far we have been working mostly with strings, and some numbers. Python can recognize different types of values. Python will recognize and treat the values `'Hello World'` and `23` differently. How can we tell how Python sees these values? There is a command `type` that reveals it. Try in the shell:

```python
type('Hello World')
```

*user runs line in the shell, reaches next Step*

#### Step: `int`

As you see Python reports that `type('Hello World')` is `<class 'str'>`. Don't worry about `class` for now. Python is telling us that the type of a string is `str`, which is short for *string*. Let's try to discover a few more types:

```python
type(23)
```

*user runs line in the shell, reaches next Step*

#### Step: `float`

Now Python tells us that `type(23)` is `<class 'int'>`. Here `int` stands for *integer*, which is one of the basic number types. What would happen if the number had a decimal point? Try:

```python
type(23.4)
```

*user runs line in the shell, reaches next Step*

#### Step: `bool`

This time it's `<class 'float'>`. Here `float` is short for *floating point number*. Don't worry about the precise meaning of this technical term for now, it's enough to know that it's about numbers with decimal points.

Interesting! So number values with and without decimal points are considered as different types by Python.

[...]

PR #108, opened 2020-12-14
added Page 1 of Chapter 11 Tic tac toe project (#3 #92)

spamegg1:
see #3 and #92

### DIFF
--- a/backend/main/chapters/c11_tic_tac_toe_project.py
+++ b/backend/main/chapters/c11_tic_tac_toe_project.py
@@ -7,11 +7,15 @@ from main.text import ExerciseStep, Page, MessageStep
 
 
 class IntroducingTicTacToe(Page):
-    title = "Introduction to the project, and the `winner` function"
+    title = "Checking the board for winners"
 
     class intro_row_winner(ExerciseStep):
         """
-In this large project you will develop a text-based interactive tic-tac-toe game to be played by 2 human players.
+You've done many short exercises solving one little problem. Now we're going to tackle a larger, more complex
+project which will really test your knowledge and require combining many smaller parts together.
+It's going to be so fun!
+
+You will develop a text-based interactive tic-tac-toe game to be played by 2 human players.
 Here is a small preview of what the finished game will look like in play:
 
      123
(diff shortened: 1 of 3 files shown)


alexmojaki:
Please also sort out https://github.com/alexmojaki/futurecoder/pull/106#issuecomment-735565744

spamegg1:
Oh right! It was on my mind but I forgot

backend/main/chapters/c11_tic_tac_toe_project.py
```diff
@@ -0,0 +1,606 @@
...
+
+        hints = """
+You need to check every row, so you'll need a loop.
+How can you check if all entries in a row are equal to each other?
+If you don't know the size of a row, you'll have to loop all the way through it.
+So you'll end up using nested loops!
+For each row, define a boolean. Then loop through that row, updating the boolean accordingly.
+You can use the first entry `row[0]` in a row to compare all the row entries to it.
+Think carefully about what the initial value of the boolean should be, and under what conditions you should change its value.
+After looping through a row, if you determined that all its entries are equal, then return `True` (ending the outer loop early).
```
alexmojaki, line 127: Need to cover this concept in the functions chapter

spamegg1: OK. Will get to it. Would `Returning Values from Functions` be the best place for it? Would one `VerbatimStep` suffice? Or should I make a new page with many steps?

spamegg1: On second thought, I should add at least 2 or 3 steps. I should show one `if-else` block example, maybe mention that `else` is unnecessary; and I should show one `for` loop example, with Snoop, so they can see the loop doesn't go all the way.

alexmojaki:
I think the early return concept needs a new page at the end of the chapter, a few steps, and a markdown draft before coding. Here are some ideas:

Explain what happens here:

```python
def foo():
    return 1
    return 2

print(foo())
```

Sometimes beginners try returning many values with something like this:

```python
def double_numbers(numbers):
    for x in numbers:
        return x * 2

assert_equal(double_numbers([1, 2, 3]), [2, 4, 6])
```

Explain why this doesn't work and that they need to create a list and append to it instead.

Emphasise that `return` ends the whole function, not just the loop. Contrast this:

```python
def foo():
    for letter in 'abcde':
        for number in range(3):
            print(f"{letter} {number}")
            if letter == 'c':
                return letter
foo()
```

with using `break` instead of `return letter`, which only breaks the inner loop.

alexmojaki:
There's still a lot of duplicated `generate_inputs` code. A higher order function will work well to reduce that.

spamegg1:
Something like this? I didn't know where else to put `generate_board`, should it go into `main/exercises.py`?

spamegg1:
The changes look good by the way, text and hints are clearer :+1:

backend/main/chapters/c11_tic_tac_toe_project.py
```diff
@@ -7,6 +7,31 @@
 from main.text import ExerciseStep, Page, MessageStep, Disallowed
 
 
+def generate_board(board_type):
+    winning, size = choice([True, False]), randint(3, 10)
+    char1, char2 = choice(ascii_uppercase), choice(ascii_uppercase)
```
alexmojaki, line 12: I don't like this style, please just assign one per line like before.

spamegg1: OK.

alexmojaki:
`generate_board` is perfect, even better than what I had in mind. No reason to move it anywhere.

spamegg1:
Well then I'll get to work on that draft

PR #111, opened 2020-12-31
added page about newlines and `format_board` to Ch 11 Tic-Tac-Toe #3

spamegg1:
Happy New Year! #3

(the "next page" link right before bonus challenge is missing for now, because the next page doesn't exist yet)

### DIFF
--- a/backend/main/chapters/c11_tic_tac_toe_project.py
+++ b/backend/main/chapters/c11_tic_tac_toe_project.py
@@ -592,7 +592,7 @@ Next we will tackle the problem of displaying the board on the screen.
 
 
 class NewlinesAndFormatBoard(Page):
-    title = "Introducing the newline character, and `format_board`"
+    title = "The newline character, `format_board`"
 
     class one_way_to_print_board(VerbatimStep):
         """
@@ -655,7 +655,7 @@ assert_equal(
 )"""
 
         def check(self):
-            return 'SyntaxError' in self.result
+            return 'SyntaxError: EOL while scanning string literal' in self.result
 
     class multi_line_strings_triple_quotes(VerbatimStep):
         """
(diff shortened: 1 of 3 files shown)


backend/main/chapters/c11_tic_tac_toe_project.py
```diff
@@ -589,3 +589,346 @@ def generate_inputs(cls):
...
+    print_board([
+        ['X', 'O', 'X'],
+        [' ', 'O', 'O'],
+        [' ', 'X', ' ']
+    ])
+
+(What's `"".join`? Google it!)
+        """
+
+        program_in_text = False
```
alexmojaki, line 615: why?

spamegg1: Hmm... I think I was under the false impression that `__program_indented__` has to be the very last thing in the class string.

spamegg1: Should be fixed in the next commit.

spamegg1:
The only way I could find to display
```python
len('\n')
```
correctly was
```
`__program__`
```
Tried other things like
```
__program__
__program_indented__
`__program_indented__`
```
, I would get either
```
len('

')
```
or
```
len(' ')
```
It still doesn't get syntax coloring though. 
Also if the user copies it from the text and pastes into the shell, it looks correct but does not pass the Step because it's actually `len('\\n')` that is copied even though it looks `len('\n')`.
Can't figure this out. It's probably something stupidly simple.

alexmojaki:
Apparently Python is weird about slashes in replacement strings when using `re`:

```python
>>> "hello bob".replace("bob", "\\n")
'hello \\n'
>>> re.sub("bob", "\\n", "hello bob")
'hello \n'
>>> re.sub("bob", "\\\\n", "hello bob")
'hello \\n'
```

The problem line was this:

```python
text = re.sub(r" *__program_indented__", indented, text, flags=re.MULTILINE)
```

alexmojaki:
I forgot that the bonus format_board would require converting int to string. Let's save that as an exercise for after they learn about types.

In the meantime I suggest changing the bonus challenge to a prettier board format:

```python
def format_board(board):
    joined_rows = []
    for row in board:
        joined_rows.append("|".join(row))
    lines = []
    for _ in board[0]:
        lines.append("-")
    line = f'\n{"+".join(lines)}\n'
    return line.join(joined_rows)

print(format_board(board = [
    ['X', 'O', 'X'],
    [' ', 'O', 'O'],
    [' ', 'X', ' ']
]))
```

Which prints

```
X|O|X
-+-+-
 |O|O
-+-+-
 |X| 
```

alexmojaki:
Merge if you're happy!

spamegg1:
Looks pretty good... I didn't think to add `TODO` for the link, good idea.
Merging...

Issue #115, opened 2021-01-31
Testing in Github actions


alexmojaki:
Need to run `pytest` under the `backend` folder, skipping `test_frontend` for now.

PR #117, opened 2021-01-31
added testing with Github Actions #115

spamegg1:
addressing #115
can be seen in action [here](https://github.com/alexmojaki/futurecoder/actions)


alexmojaki:
Great stuff! Closes #115

PR #125, opened 2021-02-04
Feature/features

ChrisjanWust:
Had to create another PR because GitHub thinks the other one is merged because I screwed a push to the origin remote up.

In any case...

@alexmojaki, I really like the text changes you made. I made one or two minor edits in a single commit. Let me know what you think. Mostly removing text that I think is redundant / could fit better in the `alt` tag.

Also, I popped some plasters on the state of our mobile site. We've got some ugly HTML hanging around which I'll need to refactor, while at the same time caring more about the mobile experience. It's not that much work, but I don't have any time at all in the next few days.


ChrisjanWust:
@spamegg1, I'm planning to merge tomorrow, if you'd want to do a quick final check.

spamegg1:
I noticed a tiny little visual inconsistency, probably not too important:
I see these horizontal white lines from the headings to the images. They directly touch the images except for "Enhanced Error Messages", so the image is too far to the left?:

page

ChrisjanWust:
Nice spotting. On what browser are you, @spamegg1? Does seem to be linked to non-webkit browsers.

ChrisjanWust:
Ah, silly me, I know what's the issue. It's some webkit targeted css that I didn't find an alternative for on non-webkit browsers. I'll have to look at it in the next PR.

ChrisjanWust:
I've made a note here: https://github.com/alexmojaki/futurecoder/issues/127

Issue #123, opened 2021-02-03
Windows compatibility


alexmojaki:
@matan-h has pointed out that futurecoder doesn't run on Windows. I don't use Windows so this is not easy for me to deal with.

I imagine that it's possible to run on Windows using Docker (right?) but that's not good enough. Developers should be able to run directly on their machine if they prefer that workflow (as I do). We also need Windows compatibility if we want to make an easy to install desktop program as described in #89.

First of all, what exactly goes wrong when running on Windows? I'm sure you can't run `setup.sh` but it should be easy to emulate the steps within: install poetry, run `poetry install`, run some manage commands. But I assume there are other problems, I haven't paid much attention to OS compatibility.

spamegg1:
I've started investigating.

It seems not possible to use Docker. Apparently my CPU does not have SLAT (Second Level Address Translation) capabilities needed to run Docker on Windows. Windows 10 won't let me install the Hyper-V Hypervisor because of this. In addition Docker needs a lot of other things, like WSL2. Also it requires Windows 10 Pro (there is some other complicated method for Windows 10 Home).

Docker is installed, but it's not really working. `docker-compose up` fails as a result with: 
```
Traceback (most recent call last):                                                                                                                              
File "site-packages\docker\api\client.py", line 205, in _retrieve_server_version                                                                              
File "site-packages\docker\api\daemon.py", line 181, in version                                                                                               
File "site-packages\docker\utils\decorators.py", line 46, in inner                                                                                            
File "site-packages\docker\api\client.py", line 228, in _get                                                                                                  
File "site-packages\requests\sessions.py", line 543, in get                                                                                                   
File "site-packages\requests\sessions.py", line 530, in request                                                                                               
File "site-packages\requests\sessions.py", line 643, in send                                                                                                  
File "site-packages\requests\adapters.py", line 449, in send                                                                                                  
File "site-packages\urllib3\connectionpool.py", line 677, in urlopen                                                                                          
File "site-packages\urllib3\connectionpool.py", line 392, in _make_request                                                                                    
File "http\client.py", line 1244, in request                                                                                                                  
File "http\client.py", line 1290, in _send_request                                                                                                            
File "http\client.py", line 1239, in endheaders                                                                                                               
File "http\client.py", line 1026, in _send_output                                                                                                             
File "http\client.py", line 966, in send                                                                                                                      
File "site-packages\docker\transport\npipeconn.py", line 32, in connect                                                                                       
[...]

PR #128, opened 2021-02-05
updated README to include Windows instructions #123

spamegg1:


spamegg1:
Eh... stupid Markdown formatting issues

spamegg1:
Nope... still looks wrong.

alexmojaki:
Do you think Windows developers need to be told how to run Powershell scripts? I'm not familiar with Powershell at all, is it not standard knowledge? If they can't figure that much out I'm not hopeful about them being able to do much else.

SET_LIMITS needs to be mentioned explicitly somewhere, users aren't supposed to run setup.ps1 every time they start the server. It has other uses anyway, like making it possible to use debuggers within the workers.

alexmojaki:
Also please check the last two steps, I think they're out of date. They shouldn't need to go straight to the login URL, right?

spamegg1:
I'm guessing that someone who is just starting to get into development on Windows would not know how to run Powershell scripts. Powershell scripting is definitely not common knowledge (I was a 20 year long Windows user). The Windows world is more sharply separated by professional/corporate vs. casual user (whereas the average Linux user is generally more capable/knowledgeable). They would start opening issues complaining why setup doesn't work... I'm just saving future headache. If you want to attract some Windows contributors there will be a lot of hand holding.

Can you tell me what the last two steps should be? `npm start` directly launches `localhost:3000` doesn't it? I usually can't follow or understand the order of what happens next.

alexmojaki:
OK I'll take care of the final steps. Happy to merge?

spamegg1:
Sure thing.

Issue #130, opened 2021-02-07
Unhandled Rejection (SecurityError) the operation is insecure (in Firefox)


spamegg1:
While developing on Firefox with `poetry` and `npm` I run into the following error every time I navigate to a page:

![linux_firefox_poetry_npm_dev](https://user-images.githubusercontent.com/4255997/107142510-a1e69a00-6940-11eb-856c-e4dca70903ee.png)

It only happens in Firefox. It is not debilitating, I can continue using the development site `localhost:3000` by clicking "X" but it is quite annoying while developing. I guess I'm the only loser in the world who still uses Firefox.

The issue is `navigator.serviceWorker.ready.then(...)` in `frontend/src/serviceWorker.js`. Specifically the `ready` part.

I also found the same issue while running the production site `futurecoder.io` on Windows with Firefox. It does not show the error page like above, but the console shows the same error `Uncaught (in promise) DOMException...` coming from the same line of code (except it's in the optimized production code `main.xxxxxxxx.chunk.js:1`).

I experimented with a few different workarounds:

- adding a `catch` clause to the `unregister` function:

  ```javascript
  export function unregister() {
    if ('serviceWorker' in navigator) {
      navigator.serviceWorker.ready.then(registration => {
        registration.unregister();
      }) // I added the line below
      .catch(err => console.log("Service worker failed to unregister"));
    }
  }
  ```

- changing `serviceWorker.unregister()` to `serviceWorker.register()` at the end of `frontend/src/index.js`

- commenting out `serviceWorker.unregister()` altogether

- deleting `serviceWorker.js` (and removing references to it in `index.js`)

### Service workers

I ended up learning about service workers. First of all they are OPTIONAL. From the top of `serviceWorker.js`:

```javascript
// This optional code is used to register a service worker.
// register() is not called by default.

// This lets the app load faster on subsequent visits in production, and gives
// it offline capabilities. However, it also means that developers (and users)
// will only see deployed updates on subsequent visits to a page, after all the
// existing tabs open on the page have been closed, since previously cached
// resources are updated in the background.

// To learn more about the benefits of this model and instructions on how to
// opt-in, read https://bit.ly/CRA-PWA
```

So unless `unregister()` is switched to `register()` the file is doing nothing, just sitting there. I confirmed this by looking at the "Applications" tab in the Developer Console. The "Service Workers" section shows nothing, so clearly `futurecoder.io` is not using them.

***What do service workers do exactly?***

[...]

PR #131, opened 2021-02-07
removed /frontend/src/serviceWorker.js #130

spamegg1:
Want to check out the deployment for console error messages, to see if `Uncaught (in promise)...` errors disappear or not.
Do not merge.


spamegg1:
Yep, the `(in promise)` errors from `main.xxxxxxxx.chunk.js:1` are gone. Same result as I have in the development environment.

PR #136, opened 2021-02-16
keep traceback tooltip popup inside the window (fixes #107)

### LINKED ISSUE #107: Information from traceback not always visible
The display can be "above" the top of the page:

image

spamegg1:
Here is a screenshot:
popup

### DIFF
--- a/frontend/src/shell/TerminalMessage.jsx
+++ b/frontend/src/shell/TerminalMessage.jsx
@@ -70,6 +70,7 @@ const Tracebacks = ({tracebacks, codeSource}) => {
                 </span>
               }
               position="top right"
+              keepTooltipInside={true}
               on={['hover', 'focus']}
               arrow={false}
               contentStyle={{


spamegg1:
Deployment looks good...
pop

alexmojaki:
This looks great, but now that you've found `keepTooltipInside`, there's not much reason to use `position="top right"`. It was the safest way to keep the tooltip inside, but it covers the code and traceback. can you try `bottom left` and see what happens? It probably won't do that in most situations when the editor is unlocked, but still.

spamegg1:
I've tried all possible positions with `keepTooltipInside`:
 "top right" | "top left" | "top center" | "right top" | "right center" | "right bottom" | "bottom left" | "bottom center" | "bottom right" | "left top" | "left center" | "left bottom" | "center center" 

They are all overridden by `keepTooltipInside`, which acts like "top right" except when it goes out the top, in which case it acts like "bottom right". This includes when the editor is unlocked.

alexmojaki:
Ok. Thanks for solving this and for checking the position!

PR #144, opened 2021-02-23
update tracebacks.py

matan-h:


spamegg1:
@matan-h Looks like `test_steps` is failing. Check out the `Testing/test_steps` part here: https://github.com/alexmojaki/futurecoder/blob/master/how_to_contribute.md
In particular, with `poetry shell` active in the backend folder, you'll have to export the environment variable `FIX_TESTS=1` and run `pytest -k test_steps`.

backend/main/workers/tracebacks.py
```diff
@@ -42,17 +42,27 @@ def didyoumean_suggestions(e) -> List[str]:
         log.exception("Failed to get didyoumean suggestions")
         return []
 
+def get_generic_explanation(e):
+    ft = FriendlyTraceback(type(e), e, e.__traceback__)
+    ft.compile_info()
```
alexmojaki, line 47: I need to look into it, but the name implies to me that this puts together lots of info, some of which I might not need. Does it? Either way, please refactor to only call this once.

PR #150, opened 2021-03-22
Fix friendly tracebacks

matan-h:


backend/main/workers/tracebacks.py
```diff
@@ -81,8 +85,8 @@ def print_friendly_syntax_error(e):
 {''.join(lines).rstrip()}
 at line {e.lineno}
 
-{friendly_generic(e)}
-{friendly_syntax_cause(e)}""",
```
alexmojaki, line 85: We still want the friendly cause of the syntax error

backend/main/workers/tracebacks.py
```diff
@@ -43,17 +48,19 @@ def didyoumean_suggestions(e) -> List[str]:
...
         return ""
 
 
-def _friendly_cause(e, setter):
-    info = {"message": get_message(type(e).__name__, e), "generic": ""}
+def _friendly_cause(fr:FriendlyTraceback,setter):
     try:
+        fr.assign_cause()
+        info = {"message": fr.info['message'], "generic": ""}
         setter(info)
```
alexmojaki, line 64: this is no longer relevant

backend/main/workers/tracebacks.py
```diff
@@ -32,6 +31,12 @@
 
 log = logging.getLogger(__name__)
 
+def get_friendly_object(e):
+    try:
+        fr = FriendlyTraceback(type(e), e, e.__traceback__)
+        return fr
+    except (Exception, SystemExit):
+        log.exception("Failed to build FriendlyTraceback object")
```
alexmojaki, line 39: This returns None and then causes another exception later. Don't call the other friendly functions when this happens.

PR #152, opened 2021-04-02
Re-style toc page

pwwang:
Make it something look like:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,6 @@ backend/main/tests/error_screenshot.png
 # temporary local files
 tmp/
 videos/media/
+
+# vscode's settings
+.vscode/
--- a/backend/main/static/css/main.scss
+++ b/backend/main/static/css/main.scss
@@ -1,6 +1,3 @@
-.toc {
-  background: white ;
-}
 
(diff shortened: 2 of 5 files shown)


alexmojaki:
It looks nice, but wasn't part of the goal to make something less bright? It's even brighter than before, definitely a bit of a shock after the dark home page.

Do you think the instructions should be kept? I think they can probably be removed, except for the "You were last on the page" case.

pwwang:
Yeah, haven't noticed the transition from the dark homepage. Will try to make it darker.

Possible to make the instructions shorter?

pwwang:
How about this:

image

alexmojaki:
I think it's good! I'm going to merge now because it's already a definite improvement.

I'm not sure what to do about the instructions, it's actually fine as it is but I'm open to ideas.

"You were last on the page" looks a bit weird, is that intentional?

Screenshot from 2021-04-03 13-14-17

The font size for the page links seems a bit small?

We now have three different dark backgrounds for the home page, ToC, and login. Maybe we need some consistency?

@spamegg1 @ChrisjanWust any thoughts?

PR #153, opened 2021-04-05
Fix "You were last on page" and increase the link font-size on toc page

pwwang:


alexmojaki:
Thanks. I wasn't quite sure about the right-aligned text, but it's growing on me.

Issue #41, opened 2020-08-08
Allow starting the course without an account


alexmojaki:
To make the site more appealing, people should be able to start using it without signing up with an email and password.

Here's how this should work:

1. When someone views the course without logging in, it currently hits this code:
 https://github.com/alexmojaki/futurecoder/blob/1f87b99ba1c017641a1d477bef4334da2c5e71a9/backend/main/views.py#L143-L146

   This ultimately tells the JS to redirect to the login page. Instead of `return {}`, it should create a new 'guest' account with a dummy email and password and login as that user (which sets a cookie in the browser or something so the login lasts for a while). Then the user can keep browsing and it will keep track of their progress.
2. Inform users with a 'guest' account that they need to sign up. Above the editor/shell, just left of the hamburger menu, show signup and login buttons. For logged in users, show their email instead of hiding it in the menu.
3. When giving feedback, add an optional field to enter an email (usually this is filled in automatically by looking at the logged in user).

Optional extensions after the above is merged:

- If the user signs up, preserve their progress and other data. Either copy data from the old user object into the newly created one, or don't create a new user at all and just set the email and password upon signup. Note that the site uses django-allauth to manage users, you will probably need to interface with that.
- After completing each chapter show a popup suggesting that they create an account to keep track of their progress. Link to the signup page. The user should be able to ignore this if they want, i.e. close the popup and continue.
- Add a "don't remind me again" checkbox to the above popup.
- In the very first page, insert some text about signing up within the content before talking about the shell. The text should only show to guest users.
- Make the "start coding" button go straight to a course page instead of the table of contents. For logged in users, go to the page they were last on as linked in the TOC. For guest users, go to the first page. Add some text at the beginning of the first page pointing to the TOC in the menu, and mention that more features are unlocked in later chapters.

alexmojaki:
Currently working on switching to Firebase instead of Django. In the process I will solve this issue, just not quite like described above.

PR #157, opened 2021-05-03
Pyodide

alexmojaki:
Not quite ready yet, but I'm at least confident that this is going to work. In short, futurecoder will run Python code in the browser, which will drastically reduce cost and scalability issues.


alexmojaki:
@spamegg1 @ChrisjanWust @matan-h check it out! All the user code is now run in the browser, not the server. Once everything is loaded you can even disconnect your internet and continue the course! This completely changes the roadmap for futurecoder and gets rid of a lot of problems. No more SET_LIMITS or complicated threads and processes - stuff that especially caused a lot of trouble for @matan-h recently.

The big caveats are:

1. It takes a while to download and initialise everything. On my computer it's a few seconds, but it'll depend on your internet and CPU. I'm curious to hear other people's experiences.
2. input() doesn't work yet, but that should be solvable.
3. Browser support is limited, for example Safari probably won't work.

pwwang:
Exciting!
Does that mean we don't need any backend thing anymore?
Just wondering what takes so long to load? Looks like [Pyodide's link](https://pyodide.org/en/stable/console.html) is pretty fast.

A heroku link to share for me or others to try?

spamegg1:
Awesome! Doing things client-side is a great advantage. (Reminds me of my old number theory project using Brython)
@alexmojaki Congrats on such a huge milestone. `Files changed: 57` Yikes! Must have been difficult. Gets me motivated to continue with my drafts!

On Firefox: opening a page from TOC:
```
Python initialization complete pyodide.asm.js:9:236650
loadPyodide: 12642.38ms - timer ended Worker.5137703f.worker.js:1:13858
Python core ready! pyodide.asm.js:9:236650
load_package_buffer(buffer): 2194.35ms - timer ended Worker.5137703f.worker.js:1:14662
```
My internet is pretty garbage (1MB/s) so keep that in mind.

There seems to be a short 1-2 second hang after a predicted output choice exercise and the confetti animation.

@pwwang Click on "View Deployment"

pwwang:
I didn't feel any delay in loading, strikingly. All worked smoothly for me.

Sounds good?
image

alexmojaki:
> Exciting!
> Does that mean we don't need any backend thing anymore?

Yes, the long term plan is:

1. Make it usable completely offline, as long as static assets have been downloaded.
2. Use something like Cloud Firestore or some serverless BaaS thing (still learning about this) to store user data and do simple database operations. That's all the server does now, nothing complicated left.
3. Serve all static assets from a CDN.

I think that should make it basically infinitely scalable for a low price.

The codebase should still allow running Python on a server for:

1. Easier local development and testing.
2. An application that users can download and install locally with a proper Python interpreter that runs faster and can do things Pyodide can't like access the local filesystem.

> Just wondering what takes so long to load? Looks like [Pyodide's link](https://pyodide.org/en/stable/console.html) is pretty fast.

For me at least Pyodide itself takes longer to load than my code. It's not that long, but if someone has a slow computer it might be a problem.

alexmojaki:
> On Firefox: opening a page from TOC:
> 
> ```
> Python initialization complete pyodide.asm.js:9:236650
> loadPyodide: 12642.38ms - timer ended Worker.5137703f.worker.js:1:13858
> Python core ready! pyodide.asm.js:9:236650
> load_package_buffer(buffer): 2194.35ms - timer ended Worker.5137703f.worker.js:1:14662
> ```
> 
> My internet is pretty garbage (1MB/s) so keep that in mind.

I only just noticed that `loadPyodide: 12642` has 5 digits so I was wondering why you said the second time was faster. So that's a pretty decent demo of why slow internet could be a problem. Nice to see the cache works well.

> There seems to be a short 1-2 second hang after a predicted output choice exercise and the confetti animation.

That's an intentional delay that's alway been there - give the user a moment to process their correct answer before it disappears.

alexmojaki:
> I didn't feel any delay in loading, strikingly. All worked smoothly for me.

The page loads first and then the Python stuff starts loading in the background. So users won't wait long before they can see something and start interacting, and they'll only notice a delay from Pyodide if they try to run code before it finishes loading. If they take their time reading text then they shouldn't notice anything, unless it takes up significant RAM or CPU resources. I don't know if that's an actual concern or not.

spamegg1:
>     2\. Use something like Cloud Firestore or some serverless BaaS thing (still learning about this) to store user data and do simple database operations. That's all the server does now, nothing complicated left.

I wonder if browser local storage or cookies can be used to keep track of which page/step the user was on last, and get rid of email/password/login completely?

alexmojaki:
> I wonder if browser local storage or cookies can be used to keep track of which page/step the user was on last, and get rid of email/password/login completely?

Client side storage will definitely be used, but you could ask the same question about many online services. I think it's pretty commonly accepted that it's important for data to be saved remotely so that users can access it from multiple devices or in case they lose their local data somehow. 

But yes, not requiring login is important, hence #41 is pinned.

pwwang:
> Yes, the long term plan is:
> 
> 1. Make it usable completely offline, as long as static assets have been downloaded.

And

> I wonder if browser local storage or cookies can be used to keep track of which page/step the user was on last, and get rid of email/password/login completely?

keep me thinking about making it completely standalone with something like Electron.

alexmojaki:
Like I said, a local standalone version will use a server with a normal Python interpreter, not Pyodide. Otherwise there's not much point.

ChrisjanWust:
Super cool that this is possible. 

Re:

> The codebase should still allow running Python on a server for:
> 
> Easier local development and testing.
> An application that users can download and install locally with a proper Python interpreter that runs faster and can do things Pyodide can't like access the local filesystem.

I don't quite follow the arguments for keeping the Python server, given that user data is going to be stored on a BaaS (new word for me too :grin: ). Easier local development and testing is, at least in my mind, less important than keeping the development and production environments as similar as possible. And also something that can be improved. What features does access to the filesystem allow for?

alexmojaki:
User data will also be on the BaaS in local development. The server will essentially only be for running code. I'll probably switch it to flask, not much need for Django.

Main advantage for local development is that it reloads code easily. Currently you have to run a script which zips up all the relevant Python code, refresh the page so that it downloads the new zip, and wait for pyodide to load again. I imagine that can be automated and improved to an extent, we'll see once we actually get there.

If at some point the course teaches users how to read and write files, we can teach that with Pyodide's virtual in-memory filesystem. But I think there's value in students interacting with actual files on their own machine and experiencing 'the real thing'.

Some other things that can't be done with Pyodide, at least not 'really' (I think):

- Threads and multiprocessing
- subprocess, e.g. run a shell command
- Servers (e.g. flask/django)
- Make network requests, e.g. talk to an API or scrape a website
- Install arbitrary packages from pypi

But all of this is looking very far ahead.

Then there's the issue of speed. Pure Python code runs about 10x slower in the browser. Most of the time this won't really matter, but it might become an issue for slow computers, complicated user code, or birdseye/snoop. Again, we'll see.

alexmojaki:
:confetti_ball:

PR #159, opened 2021-05-29
Make menu bar in top left with user and toc link

alexmojaki:


alexmojaki:
This is based on conversations with @ChrisjanWust and @pwwang. The plan is that the user indicator can be replaced by a button to login/sign up when the user is not signed in. So while it's not a big deal right now, it will be.

It's also nice to draw some more attention to the table of contents. We can also move other items from the menu dropdown into the top bar if we think they're important enough, or put other information in there, such as the current chapter, page, and step number.

I know it doesn't look great, but does it look OK?

PR #161, opened 2021-06-06
No django

### LINKED ISSUE #83: Proper navigation (back button, URLs) in React
Currently you can refer to a specific page with a URL like https://futurecoder.io/course/?page=FunctionsAndMethodsForLists

This is very useful, but the URL doesn't change when going to the next or previous page. This is annoying because refreshing brings you back to the original page.

It would also be nice if the navigation history was populated so that pressing back in the browser worked intuitively.

alexmojaki:
Supersedes #160 but I'm leaving that one open for now because I haven't set up github actions or PR previews here yet.

This is it! No more servers! Currently hosted here: https://futurecoder-io.web.app/

Also closes #83 

Remaining things to do:

- [x] Update documentation for running and developing. There's no more docker.
- [x] Github actions
- [x] PR previews from firebase
- [x] Make a new GitHub API key and put it somewhere more secret looking because when I pushed the last one I instantly got an email from GitHub that they'd automatically revoked it for my safety.
- [x] Update the homepage a bit:
  - [x] Remove django icon, maybe add firebase and pyodide
  - [x] Remove/update paragraph about donations, especially "This site currently runs on low-cost hardware and cannot handle a large number of users simultaneously."
  - [x] Email list, temporarily removed
- [x] Get new Sentry plan separate from heroku
- [x] Migrate user data to firebase database
- [x] Switch domain from heroku and cloudflare to firebase


alexmojaki:
:confetti_ball:

PR #162, opened 2021-07-03
Add page with input() and framework for exercises with input()

alexmojaki:
@spamegg1 how's this?

Code needs a lot of cleanup, plus there should be some documentation on how to write these kinds of steps.


spamegg1:
The API looks very intuitive and easy to use!

alexmojaki:
@spamegg1 are you able to review? Do you think I can merge?

PR #173, opened 2021-08-29
Automatically load packages from imports

alexmojaki:
This only works for pyodide packages like numpy, not arbitrary packages on pypi. To install from pypi, `import micropip; micropip.install("package_name")`. Then wait a bit before trying to import the package, maybe check the dev console. I've asked in https://github.com/pyodide/pyodide/issues/1470#issuecomment-907815342 how to easily automate this in the same way.

Only some packages can be installed like this - in particular there has to be a pure python `.whl` on pypi.

### DIFF
--- a/core/workers/worker.py
+++ b/core/workers/worker.py
@@ -90,7 +90,13 @@ def check_entry_catch_internal_errors(entry, input_callback, result_callback):
 
 def check_entry(entry, input_callback, result_callback):
     if hasattr(entry, "to_py"):
+        import js
         entry = entry.to_py()
+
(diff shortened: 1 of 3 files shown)


alexmojaki:
Example script to try out:

```python
import pandas
import micropip
micropip.install("icecream")
print(pandas.DataFrame)
import icecream
icecream.ic(1+2)
```

This takes a while to run as pandas and micropip are installed. `micropip.install("icecream")` runs asynchronously so `import icecream` fails the first time. Run the script a second time and it should all work quickly.

PR #174, opened 2021-08-29
Full IDE mode: no course text, screen is filled with shell and editor.

### LINKED ISSUE #118: Full screen editor mode, or suppressing messages
This idea came out of a discussion with @matan-h about an issue we've both noticed. If the user wants to play and experiment with some code not related to the current step, they can technically do so at any time, but the system assumes they are trying to solve the current step and may display messages to help them with that. We want users to feel free to try things out, whether it's something completely random or it's tinkering with code from the previous step to refine their understanding of it. It can suck to be doing that and suddenly get a message "You need to define a function named foo".

We could have a little toggle switch thing to turn off these messages, or an entirely separate page which has the coding tools but no course material, but I don't know if these would be used much.

So I'm thinking it might be nice to have a 'maximise' button which expands the IDE on the right half to fill the whole screen: the editor fills up the left half, and the shell fills the right half. This would probably be a nicer editing experience overall, and as a side effect it would allow the user to feel detached from the current course material. They don't feel pressure to do what the text tells them, and messages would not show in this mode. Users can easily switch back and forth between the two modes without changing what's in the editor/shell.

The downsides of not showing messages are:

- Maybe the user is actually trying to solve the current step
- Some messages can still be helpful even if you're not solving a specific step, such as linting.

Possible solutions:

- Show some little bubble or other indicator that messages exist if the user wants to expand them, but this largely takes as back to square one.
- Show information such as linting in the shell output rather than messages.

Other questions:

- If messages don't show while in full screen, are they still created so that they can be seen when you go back to normal mode?
- What happens if the user completes a step? Does it automatically return to normal mode?

alexmojaki:
Happens at path #ide
Button in top right to switch between modes
Closes #118

### DIFF
--- a/core/workers/worker.py
+++ b/core/workers/worker.py
@@ -1,4 +1,5 @@
 import builtins
+import importlib
 import linecache
 import logging
 import sys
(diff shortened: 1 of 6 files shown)


alexmojaki:
@matan-h check it out, this is based on your request.

Also @spamegg1 @pwwang @ChrisjanWust I think you'll like this. Makes futurecoder useful to anyone, not just beginners. Good alternative to repl.it for anyone wanting a quick place to write and debug code. Also good for teachers who want the IDE but not the course.

I'm thinking about changing `>>> Start coding` on the home page to two buttons, maybe "`>>> Start learning` or `>>> Just code`"

ChrisjanWust:
Pretty flippen sweet!

Issue #86, opened 2020-10-18
Animating the appearance of new steps in the course


alexmojaki:
Each time the user passes a step, a new step appears in the text on the left. Right now this is easy to miss, especially since their attention is usually focused on the right, particularly after answering an output prediction question and/or if the new step is short. It should be made more obvious so that users know they've succeeded. I'm thinking some animation would be helpful here. Perhaps a 'wipe' transition (like the output prediction questions) and/or a change in background (e.g. if the new step started out a pale yellow for a few seconds). Suggestions welcome!

alexmojaki:
Done in #205

PR #205, opened 2021-11-01
Animating the appearance of new steps

vokounigor:
Hey Alex!
I like the idea of this project so I decided to help out a bit. I saw the issue #86 and added some animations when moving further in the course. I used the light yellow color (as suggested) and liked the way it looked as it reminded me for some reason on Golangs' 'A tour of go'. I also added an animation when moving through pages (but only forward), as I like having a visual indicator for that.


frontend/src/css/main.scss
```diff
@@ -387,3 +387,25 @@
...
+  to {
+    background-color: inherit;
+  }
+}
+
+// Body transition
+
+@keyframes bodyTransition {
+  from {
+    opacity: 0.5;
```
alexmojaki, line 406: I'm not against an animation for changing the page, but I just tried this and it feels weird. Looks like a flash of lightning or something.

frontend/src/book/store.js
```diff
@@ -131,16 +131,34 @@ export const setPageIndex = (pageIndex) => {
...
   setUserStateAndDatabase(["pagesProgress", localState.user.pageSlug, "step_name"], step.name);
+  if (delta > 0) {
+    animateNextStep(stepIndex);
+  }
+};
+
+const animateNextStep = (stepIndex) => {
+  const div = document.getElementById(`step-text-${stepIndex}`);
+  div.style.animation = 'nextStepTransition 5s ease-out';
+  setTimeout(() => div.style.animation = '', 1000);
```
alexmojaki, line 161: This is close, but in terms of appearance:

1. It disappears quite abruptly, I think the setTimeout after 1 second is too short when the transition time is 5s. I played around a bit and liked 4s transition and 1.5s setTimeout. How's that seem to you?
2. There's a white gap above the text. Can you make the color fill the whole space between horizontal lines?

frontend/src/book/store.js
```diff
@@ -131,16 +131,34 @@ export const setPageIndex = (pageIndex) => {
...
   }
+  
   setUserStateAndDatabase(["pagesProgress", localState.user.pageSlug, "step_name"], step.name);
+  if (delta > 0) {
+    animateNextStep(stepIndex);
+  }
+};
+
+const animateNextStep = (stepIndex) => {
+  const div = document.getElementById(`step-text-${stepIndex}`);
```
alexmojaki, line 159: I get an error that the div is null when clicking "Skip step" (see https://github.com/alexmojaki/futurecoder#controls)

alexmojaki:
Thank you for doing this! Always exciting to get a new contributor. Plus I hate dealing with CSS (especially animations) so it's nice to have someone else to help with that.

frontend/src/css/main.scss
vokounigor: True. Now that I look at it, it does seem weird. I added now a sort of 'fade-in' effect which only affects course text (won't affect terminal this time).

frontend/src/book/store.js
vokounigor: It seems I was on another branch while testing these animations. I liked the 4s transition with 2s timeout. The <hr> element was causing the space to not be filled, so I had to change a bit how it is positioned in html. Looks better now, but do check it out.

frontend/src/book/store.js
vokounigor: Should be fixed now. When steps are skipped, the dom doesn't update at the same time, so I added a quick timeout of 1ms and it should now catch it.

frontend/src/App.js
```diff
@@ -275,22 +275,22 @@ const CourseText = (
...
+    <div className='pt-3'>
       {page.index > 0 &&
       <button className="btn btn-primary btn-sm previous-button"
               onClick={() => movePage(-1)}>
         Previous
       </button>}
       {" "}
       {page.index < Object.keys(pages).length - 1 && step_index === page.steps.length - 1 &&
-      <button className="btn btn-success next-button"
+      <button className="btn btn-success btn-sm next-button"
```
alexmojaki, line 293: I had the next button intentionally bigger than the previous button because it's probably needed more often. Do you think it looks too weird like that? If it needs to be consistent I think I'd prefer if they were both the default size, not small.

frontend/src/App.js
```diff
@@ -275,22 +275,22 @@ const CourseText = (
...
+      <div key={index} id={`step-text-${index}`} className={index > 0 ? 'pt-3' : ''}>
         <Markdown html={part.text} copyFunc={text => bookSetState("editorContent", text)}/>
-        <hr/>
+        <hr style={{ margin: '0' }}/>
       </div>
     )}
     <Messages {...{messages}}/>
 
-    <div>
+    <div className='pt-3'>
```
alexmojaki, line 285: Please add a comment explaining what pt-3 means.

alexmojaki:
The page transition looks less weird than before, but I'm still not a fan. Sorry, but I think it should just be left out.

The step transition looks nice now, but I don't think it's quite accomplishing the main goal of the issue, which is to make sure the user notices the new step when their attention is fixated on the right half of the screen. Do you think you could add something with a bit of motion?

frontend/src/App.js
vokounigor: I did notice the 'Next' button was bigger, but I didn't think it was on purpose. It does look kind of 'out of place' as it messes with the flow of the page. Both buttons being big or small would be consistent and would look better imo.

alexmojaki: OK, please remove the btn-sm from both then.

frontend/src/App.js
vokounigor: pt-3 is a shorthand that is used in most css libraries and is found somewhere in here, though I could not determine where. It is shorthand for padding-top 3 which should equal to 1 rem (16px by deafult).

alexmojaki: I'm guessing it's coming from bootstrap, please just add this description in a comment in code.

vokounigor: Yes, you are correct. Didn't realize bootstrap was used via link in index.html, I thought it was downloaded. The documentation is here, though I will add it in the coment: https://getbootstrap.com/docs/4.0/utilities/spacing/

vokounigor:
>  Do you think you could add something with a bit of motion?

Do you have any specific thing in mind when you say motion?

alexmojaki:
Either a 'wipe' into view similar to the multiple choice questions in the shell, or having it slide upwards into position.

frontend/src/css/main.scss
vokounigor: Page transitions have now been removed.

frontend/src/App.js
vokounigor: Done, both buttons are now the same height.

vokounigor:
I've added the slide 'upward into position' animation as it looks good imo. I also realized that I needed a way to track if the user is going forward or back through pages so as to not play the animation all the time. I think it looks ok now, but do check it out.

alexmojaki:
The animation looks good but the behaviour of when the animation shows is now weird. I don't expect the animation when loading the page or navigating between pages, only when advancing a step within the page. The previous method seemed right to me, I don't understand why it changed?

vokounigor:
The previous method couldn't support transitions as that is how rendering to the DOM works. If there was a transition set, an element would first be rendered and in position for a split second and then an animation would trigger which looked weird and out of place. Therefore, an animation has to be triggered once the element is created.

I like this current animation as it easily tells the user where he left off and what step he currently is on. Also moving between pages may be disorienting to some users, but this animation will point out the last step a user has completed (There is a bug here where I forgot to include only the last element, but that can be fixed).

From a UX standpoint, this seems like a good thing, but I understand if you don't like it. If you want it to only trigger the animation when a step is completed, I could look into it and configure it. I'm glad you like the animation btw :)

alexmojaki:
> The previous method couldn't support transitions as that is how rendering to the DOM works. If there was a transition set, an element would first be rendered and in position for a split second and then an animation would trigger which looked weird and out of place. Therefore, an animation has to be triggered once the element is created.

OK, I understand why things had to change. Can you remove the `slice` so that DOM elements for all steps exist from the beginning, and instead hide elements until it's time to reveal them?

> I like this current animation as it easily tells the user where he left off and what step he currently is on. Also moving between pages may be disorienting to some users, but this animation will point out the last step a user has completed

I don't think the user needs that because the most relevant step is always the last one. If you think having some indication like that is nice then I'd prefer only having the background color change in that case, and reserve the movement for the step transition. Note that the page already instantly scrolls to the last step.

vokounigor:
Hey! Sorry, I've been busy yesterday and had no time to look into this.

Great idea for rendering all steps at once and then hiding the ones that need to be taken. Animations should now trigger only when moving on to the next step.

frontend/src/book/store.js
```diff
@@ -132,17 +132,43 @@ export const setPageIndex = (pageIndex) => {
 
 export const movePage = (delta) => {
   setPageIndex(currentPage().index + delta);
+  if (delta < 0) {
+    removeStepsAnimations();
+  }
```
alexmojaki, line 137: When is this needed? I removed it and didn't see any animations when moving back.

frontend/src/App.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
```
alexmojaki, line 1: unused

frontend/src/App.js
```diff
@@ -17,6 +17,7 @@ import {
   setDeveloperMode,
   setEditorContent,
   specialHash,
+  userMoveDirection,
```
alexmojaki, line 20: .

frontend/src/book/store.js
```diff
@@ -132,17 +132,43 @@ export const setPageIndex = (pageIndex) => {
...
+    animateStep(stepIndex);
+  }
   setUserStateAndDatabase(["pagesProgress", localState.user.pageSlug, "step_name"], step.name);
 };
 
+const animateStep = (stepIndex) => {
+  const stepDiv = document.getElementById(`step-text-${stepIndex}`);
+  if (!stepDiv) {
+    return;
+  }
```
alexmojaki, line 168: I think this is no longer needed thanks to the new system

frontend/src/book/store.js
vokounigor: It is needed to cover a case when a user moves quickly back after submitting a step. Not very likely to happen, but in case it does, this will prevent the animation from playing. Here's how it looks when moving back fast without this: https://ibb.co/4W09VhR

frontend/src/App.js
vokounigor: Removed

frontend/src/App.js
vokounigor: Removed

frontend/src/book/store.js
vokounigor: Removed

frontend/src/book/store.js
alexmojaki: OK, I understand. Very meticulous! I think I'd prefer to remove this and keep the code simple. I'm fine with a harmless animation showing on extremely rare occasions.

vokounigor: Sure thing. Removed.

alexmojaki:
Thank you! Both the animation and the code look nice now.

How do you feel about taking on #207?

Issue #207, opened 2021-11-01
Fix dialog and flow when signing up or logging in.


alexmojaki:
After signing up and submitting an email and password, the modal dialog collapses to a thin white rectangle. Clicking outside removes it but the login/signup button remains. Refreshing shows the email correctly.

Logging in also shows the thin white rectangle which is not great, but at least it disappears automatically, although a bit too slowly.

alexmojaki:
Fixed in #217

PR #217, opened 2021-11-08
Fix dialog when signing up / logging in.

vokounigor:
Hey Alex! Related to #207.

Popup should now close properly and I also added a page refresh on sign up. The flow seems better now, but you'll be the judge of that.

I've also moved login info into a new component as it's much more readable imo (wouldn't be a bad idea to move other components from App.js into their own respective files).

I've also added a .editorconfig file so that people who join don't need to configure their editors (tabs, spaces, new lines etc.). With intelliJ tools it works out of the box, but for vscode an extension is needed. I've left it for you to configure it for python files, and I've set js and css files based on what I've seen in the code.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,7 +13,7 @@ indent_size = 2
 trim_trailing_whitespace = true
 
 # Config for Python (uncomment what you need)
-# [*.py]
-# indent_style = space
-# indent_size = 4
-# trim_trailing_whitespace = true
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# For VSCode install the extension: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig
+# top-most EditorConfig file
+root = true
+
+# Config for all files
+[*]
+insert_final_newline = true
+
+# Config for JavaScript files and CSS
+[*.{js,jsx,css,scss,sass}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
+# Config for Python (uncomment what you need)
+# [*.py]
+# indent_style = space
+# indent_size = 4
+# trim_trailing_whitespace = true
--- a/frontend/src/components/HeaderLoginInfo.jsx
+++ b/frontend/src/components/HeaderLoginInfo.jsx
@@ -2,7 +2,7 @@ import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
 import {faUser} from "@fortawesome/free-solid-svg-icons";
 import StyledFirebaseAuth from "react-firebaseui/StyledFirebaseAuth";
 import firebase from "firebase";
-import {updateDatabase} from "../book/store";
+import {updateDatabase, updateUserData} from "../book/store";
 import Popup from "reactjs-popup";
 
 const HeaderLoginInfo = ({ email }) => {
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+[*.{js,jsx,css,scss,sass}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
+# Config for Python (uncomment what you need)
+# [*.py]
+# indent_style = space
+# indent_size = 4
+# trim_trailing_whitespace = true
```
alexmojaki, line 19: This is all fine, you can uncomment

frontend/src/components/HeaderLoginInfo.jsx
```diff
@@ -0,0 +1,61 @@
...
+            // firebase.auth.GoogleAuthProvider.PROVIDER_ID,
+            // firebase.auth.FacebookAuthProvider.PROVIDER_ID,
+            // firebase.auth.GithubAuthProvider.PROVIDER_ID,
+          ],
+          autoUpgradeAnonymousUsers: true,
+          callbacks: {
+            // Avoid redirects after sign-in.
+            signInSuccessWithAuthResult: () => {
+              close();
+              window.location.reload();
```
alexmojaki, line 38: I feel like this is supposed to be possible without reloading the page, but I don't want you to waste a lot of time trying to figure out how. If we need to reload, just remove that comment.

alexmojaki: One issue with reloading is that `loadUser` is supposed to merge the local redux state with data from firebase, so logging in to an existing account preserves progress you made while anonymous. `loadUserAndPages` needs to complete and save the merged data back to the database before the page reloads and loses that state. That sort of seems to be happening but I can't see why and I'm worried it only seems to work locally where it doesn't need to wait for the internet.

vokounigor: This component should update once a users' state changes, so I was confused as to why it wasn't happening (no updates were detected in redux) and added a refresh for this to happen. Upon further inspection, it seems that
```
firebase.auth().onAuthStateChanged(...)
```
in _src/book/store.js_ triggers when user state is changed, but only on sign in, sign out or token change (expiry or password change). I've looked into it and it seems that the best solution is to replace this method with
```
firebase.auth().onIdTokenChanged(...)
```
which takes the same parameters and triggers upon sign in, sign out and all token refresh events (meaning it'll trigger on sign ups too!). I've tried it and it seems to be working fine, but It'd be best if you looked into it too.

.editorconfig
vokounigor: Done

frontend/src/components/HeaderLoginInfo.jsx
alexmojaki: It sounds like onIdTokenChanged also just gets called periodically when the token is refreshed without any user action. I don't like the sound of that.

From experimenting, it seems like signInSuccessWithAuthResult only gets called when signing up with a new account, not signing into an existing account (I assume that goes through signInFailure). And in that case onAuthStateChanged never gets called. So what if we explicitly made onAuthStateChanged and signInSuccessWithAuthResult both call the same function (which in turn calls loadUser)?

vokounigor: Reverted to onAuthStateChanged and explicitly updated store with user data on sign in success.

alexmojaki:
Looks like it works great, thank you!

Issue #219, opened 2021-11-09
Too many hints


alexmojaki:
I got some interesting feedback from a user: https://www.reddit.com/r/learnprogramming/comments/qjs3wh/i_built_futurecoder_a_100_free_and_interactive/hjsq5lc/?context=6

The essential bit:

> I revealed a hint, tried fiddling with what I had keeping the hint in mind, revealed another hint, etc. As for unshuffling the scrambled solution, no. By the time I got to those bits I was so frustrated that I just clicked through.

> In that particular challenge there were way too many [hints]. I think if you condensed them into a central idea instead of piece by piece it would have been better.

My response:

> In general I tried to make hints as gradual as possible to avoid revealing more information than needed. I see now I was thinking too simplistically and too many hints can backfire.

> What do you think of an indicator like "Shown 5 of 20 hints" (there were actually 20 hints in that exercise) so that users know what to expect and don't spend too long thinking about each hint and getting frustrated?

Their response:

> If you don't want to condense some then yes, an indicator would absolutely help.

> I generally like the gradual reveal of the hints because it allows me to think about how to get the solution without getting the answer. Sometimes I'll use to make sure I'm on the right track, which I think is what backfired on me in that exercise.

I can see now that I may have gone overboard with hints in some places. Going through the exercises and trimming them would probably be good. In the meantime adding an indicator about the number of hints seems like a good step. @vokounigor what do you think of this? Want to add an indicator? Should be very simple.

alexmojaki:
Indicator was added in #224. Keeping this open to consider reducing the number of hints in some exercises.

colevandersWands:
Maybe part of the problem is having users with different levels of experience and intuition.  IMO it would be impossible to get the hints right for everyone in on try, so what if there were an option to choose your "level" next to the hints button?  

Picking "easy" would queue up all the hints and extra support, picking "hard" would queue up fewer hints with less support.  and a few levels in between.  This would mean writing more hints, but I'm assuming that after putting in the effort to simplify the hints you already have an idea what the levels could be.  

The unknown for me is how hard it would be to implement the feature and if it would be worth the effort.

alexmojaki:
I expect 'advanced' users to not need hints at all. And when a user was struggling (like in the original feedback) it seems that more hints can make things *worse* as information is revealed too slowly.

I'm going to close this for now, I actually don't see myself taking further action on it unless it comes up again. I think the indicator should make a big difference and I think some users will still appreciate having many small hints.

PR #222, opened 2021-11-10
Added display for num of hints. Fixed bug for resetting hint count.

vokounigor:
Regarding #219.

Hint popup should now show current hint you're on as well as number of total hints for the current step.

I've also found a bug that hints would not reset when moving through steps, so if you used 8 / 8 hints and then moved onto the next step that had 10 hints, 8 out of 10 hints would already be revealed. This should fix that bug as well.

### DIFF
--- a/frontend/src/Hints.js
+++ b/frontend/src/Hints.js
@@ -40,7 +40,7 @@ export const HintsPopup = ({hints, numHints, requestingSolution, solution}) => {
 const Hints = ({hints, numHints, requestingSolution, solution}) =>
   <div className="markdown-body">
     {hints.slice(0, numHints).map((hint, index) =>
-      <div className="hint-body">
+      <div className="hint-body" key={index}>
         <div key={index} dangerouslySetInnerHTML={{__html: hint}}/>
         <hr/>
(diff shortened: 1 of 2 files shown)


alexmojaki:
You're on fire! Seems like I struck contributor gold :grin: I've sent an invite to make you a collaborator so you can make PRs from branches instead of forks and firebase can deploy previews to make things easier.

That hints bug only happens when moving directly with the developer mode buttons, right? There's some code which resets hints and several other things (e.g. solutions and output predictions) after passing a step by running the correct code. If that's still working correctly and you want to reset when clicking the buttons then might as well extract that code to reset the other stuff.

vokounigor:
Yes, you're right, it only triggers when skipping steps. I've removed it from store.js as it is working fine, and a regular user won't have access to dev mode anyway.

alexmojaki:
Developer mode isn't restricted, anyone can use it, although it feels like no one has ever discovered it on their own lol.

Can you make a new PR using a branch of the main repo instead of a fork?

vokounigor:
Sure thing. Done.

PR #224, opened 2021-11-12
Added display for number of hints.

vokounigor:


alexmojaki:
Sweet, it's nice to have these previews.

Let's try making this a little cleaner:

1. Don't show the indicator when it's shown all hints
2. That should mean that "hint(s)" can just be "hints".
3. I *think* it'd be nicer to have it directly under the button, in slightly smaller text.

vokounigor:
Previews sure are handy! Check it out now.

frontend/src/Hints.js
```diff
@@ -58,6 +58,10 @@ const Hints = ({hints, numHints, requestingSolution, solution}) =>
           />
       }
     </div>
+    {
+      numHints !== hints.length &&
+      <span className="small">Shown {numHints} of {hints.length} hints.</span>
```
alexmojaki, line 63:
```suggestion
      numHints < hints.length &&
      <span className="small">Shown {numHints} of {hints.length} hints</span>
```

I moved a step and it said "Shown 10 of 5"

alexmojaki:
This is great, thank you! What do you think about having a chat on slack?

PR #248, opened 2021-12-18
Runner refactoring

alexmojaki:


alexmojaki:
This is a giant refactoring which extracts a chunk of code into a new library https://github.com/alexmojaki/python_runner which can be used in other projects such as https://github.com/dodona-edu/papyros/issues/63#issuecomment-998912058, paves the way for more similar refactoring in the future, and also adds a couple of features/fixes, particularly:

1. Output is shown as the script is running (unless the current step has a prediction question), not only after it finishes, which is helpful e.g. when the code has an infinite loop.
2. Running code can be interrupted with a Stop button without having to run code again immediately.

These can be seen by running code like this:

```python
for x in range(10000000):
    print(x)
```

This isn't perfect. Too much output delays the effect of the Stop button. One way to see this is to run that code with Snoop and let it run for a few seconds before stopping. But fixing this turns out to be tricky and it's good enough for now.

@spamegg1 it would be great if you could test this and try breaking things. Everything is potentially affected and it's a bit scary. On the plus side this should be the start of making the codebase more robust and better tested automatically - see the high test coverage in the python_runner package.

PR #266, opened 2022-01-06
More translation infrastructure

alexmojaki:


alexmojaki:
This is now a real PR. The preview for the spanish translation is now at https://futurecoder-io--es-latam-t6zdvapg.web.app/course/

PR #269, opened 2022-01-06
Make Firebase work offline

saviosg:
With these changes, it's now possible to run it locally without ever needing an internet connection. When the user connects to the internet for the first time, they will login normally.

If the user previously connected to the internet, their previously saved data will be loaded. However, progress that was made while offline will not be saved.

I haven't tested, but I suppose this also makes futurecoder work in China.

### DIFF
--- a/frontend/src/book/store.js
+++ b/frontend/src/book/store.js
@@ -31,6 +31,12 @@ if (isProduction) {
   firebaseAnalytics = firebase.analytics(firebaseApp);
 }
 
+const isOffline = new Promise((resolve) => {
+  axios.get(`${databaseUrl}/.json`).catch((e) => {
+    resolve(!e.response);
+  });
+});
+
 const initialState = {
   error: null,
   route: "main",
@@ -65,6 +71,7 @@ const initialState = {
   },
   processing: false,
   running: false,
+  offline: false,
   numHints: 0,
   editorContent: "",
   messages: [],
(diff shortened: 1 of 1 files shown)


alexmojaki:
Not saving progress while offline is a problem. If my connection has a temporary problem while trying to load the user, I'd prefer for the page to tell me there's a problem and prompt me to refresh and maybe look for a better connection. I'd be annoyed if I spent a lot of time working through the course and then found I had to repeat that work when I came back later. In particular this will be especially annoying in China.

saviosg:
Yes, I share the concerns you raised. Users wouldn't be happy if they lost all progress when they went online, and all this data handling under the hood should be transparent to them.

We could implement offline persistence using something like [localForage](https://github.com/localForage/localForage).

alexmojaki:
localForage is already a dependency for other reasons and using it is the natural option if we decide to implement this manually. However I'm still nervous about doing that ourselves, especially the messiness of combining online and offline data when the connection comes and goes. As I said on Slack, the firebase client *should* handle this itself and would probably do it better than we would. I think the best option overall is to host a proxy at a subdomain like db.futurecoder.io which forwards requests to firebase. Then we could use the real firebase client, have good offline support, and even have full online support for China. But I haven't had time to look into this properly.

saviosg:
A proxy sounds like a good idea. So if I get this correctly, the headers needed for cross-origin isolation can be set by it.

alexmojaki:
I think so, yes

saviosg:
Alright! Closing this PR then.

alexmojaki:
I've looked into it, and for reasons I don't understand, pointing the firebase client at a proxy seems to be almost impossible. So that's not going to happen. In the long term I think this means I will need to switch to an alternative service, but that's a big undertaking. For now I think we should go forward with this. If the user is offline:

1. Make it visually obvious, e.g. replace 'Login/Signup' with a red 'OFFLINE' or something.
2. Use localForage instead of firebase for user data. For now the two can be completely separate, we can later think about merging them when the user comes back online.

alexmojaki:
Closing in favour of https://github.com/alexmojaki/futurecoder/pull/341

Issue #232, opened 2021-11-25
Make test_frontend more reliable and maybe faster


alexmojaki:
test_frontend is often failing in github actions, but not consistently. Rerunning sometimes makes it pass. So there's some race condition timing issue. Adding a sleep would probably help but in general the test is very slow and not well written, it'd be great to make it smarter and more efficient about waiting until certain things are ready.

alexmojaki:
Seems reliable now

PR #270, opened 2022-01-09
expanding explanation of  `range()`

### LINKED ISSUE #261: Expanding explanation of range()
Feedback from a user:

> For me, one lesson, https://futurecoder.io/course/#GettingElementsAtPosition , goes way too fast. It might be just me and others don't find it as difficult. I'm struggling to really understand what `range()` is doing in the loop, and I don't really understand why `print(i)` works. Again, could just be me and I'm sure if I keep doing it I'll figure it out, but it felt faster than other lessons and the only time I've gotten frustrated.

I made a proposal to improve this in https://futurecoder.slack.com/archives/C01B17RT9A5/p1640944937005400. Here's a copy, but the markdown didn't transfer and I only fixed part of it before deciding to link instead.

`...<first two steps>...`

-----

There you go. `words[4]` and beyond don't exist, so trying that will give you an error.

That first program is a bit repetitive. Let's improve it with a list and a loop!
  
    words = ['This', 'is', 'a', 'list']
    indices = [0, 1, 2, 3]
    
    for index in indices:
        print(index)
        print(words[index])

------

That's a bit better, but writing out `[0, 1, 2, ...]` isn't great, especially if it gets long.
There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`,
so indices = range(4)

------

As you can see, the result is the same. Try this:

__copyable__
indices = range(4)

print(indices[0])
print(indices[1])
print(indices[2])
print(indices[3])

What do you expect the output to be?

    0
    1
    2
    3

    1
    2
    3
    4

    [0]
    [1]
    [2]
    [3]

    This
    is
    a
    list

    (etc...)

---

Now try indices[4] in the shell.

What do you expect the output to be?

- 0
- 1
- 2
- 3
- 4
- Error

---

range(4) is the same thing as [0, 1, 2, 3] ... almost. Try range(4) in the shell.

---

That's probably a bit surprising. If you're curious, the 0 represents the start of the range.
0 is the default start, so range(4) is equal to range(0, 4).
4 is the end of the range, but the end is always excluded, so the last value is 3.
If you're confused now, don't worry about it.

There's a good reason for why range(4) is not actually a list - it makes programs faster and more efficient.
It's not worth explaining that more right now.

But you can easily convert it to a list: try list(range(4)) in the shell.

What do you expect the output to be?

- range(4)
- range(0, 4)
- list(range(0, 4))
- range(0, 1, 2, 3)
- (0, 1, 2, 3)
- [0, 1, 2, 3]

---

That's just a demonstration to let you see a range in a more familiar form.
You should almost never actually do that.

If you're feeling overwhelmed, don't worry! All you need to know is that range(n)
is very similar to the list [0, 1, 2, ..., n - 2, n - 1].

now introduce len() and then range(len())

spamegg1:
fixes #261 
Sorry about all the commits. It just so happens that front-end tests are testing this specific page! :sob: I added 6 steps, so increased the skipped steps from 2 to 8. I guess I should take a more serious look at #232 but I don't want to do that for now.

### DIFF
--- a/core/chapters/c06_lists.py
+++ b/core/chapters/c06_lists.py
@@ -427,6 +427,8 @@ That first program is a bit repetitive. Let's improve it with a list and a loop!
 __program_indented__
         """
 
+        auto_translate_program = False
+
         def program(self):
             words = ['This', 'is', 'a', 'list']
             indices = [0, 1, 2, 3]
@@ -438,7 +440,8 @@ __program_indented__
     class using_range_instead_of_indices(VerbatimStep):
         """
 That's a bit better, but writing out `[0, 1, 2, ...]` isn't great, especially if it gets long.
-There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`.
+There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`,
+i.e. `indices = range(4)`.
         """
 
         program_in_text = False
@@ -457,8 +460,6 @@ As you can see, the result is the same. Try this:
 
     __copyable__
     __program_indented__
-
-What do you expect the output to be?
         """
 
         predicted_output_choices = ["""\
@@ -552,7 +553,9 @@ That's just a demonstration to let you see a range in a more familiar form.
 You should almost never actually do that.
 
 If you're feeling overwhelmed, don't worry! All you need to know is that `range(n)`
-is very similar to the list `[0, 1, 2, ..., n - 2, n - 1]`.
+is very similar to the list:
+
+    [0, 1, 2, ..., n - 2, n - 1]
 
 By the way, you can get the number of elements in a list (commonly called the *length*) using `len(words)`.
 That means that the last valid index of the list is `len(words) - 1`, so the last element is `words[len(words) - 1]`. Try these for yourself.
--- a/core/chapters/c06_lists.py
+++ b/core/chapters/c06_lists.py
@@ -419,9 +419,140 @@ This also means that the last index in this list of 4 elements is 3. What happen
         def check(self):
             return "IndexError" in self.result
 
-    class introducing_len_and_range(VerbatimStep):
+    class improve_with_list_and_loop(VerbatimStep):
         """
 There you go. `words[4]` and beyond don't exist, so trying that will give you an error.
+That first program is a bit repetitive. Let's improve it with a list and a loop!
+
+__program_indented__
+        """
+
+        def program(self):
+            words = ['This', 'is', 'a', 'list']
+            indices = [0, 1, 2, 3]
+
+            for index in indices:
+                print(index)
+                print(words[index])
+
+    class using_range_instead_of_indices(VerbatimStep):
+        """
+That's a bit better, but writing out `[0, 1, 2, ...]` isn't great, especially if it gets long.
+There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`.
+        """
+
+        program_in_text = False
+
+        def program(self):
+            words = ['This', 'is', 'a', 'list']
+            indices = range(4)
+
+            for index in indices:
+                print(index)
+                print(words[index])
+
+    class printing_the_range(VerbatimStep):
+        """
+As you can see, the result is the same. Try this:
+
+    __copyable__
+    __program_indented__
+
+What do you expect the output to be?
+        """
+
+        predicted_output_choices = ["""\
+0
+1
+2
+3
+""", """\
+1
+2
+3
+4
+""", """\
+[0]
+[1]
+[2]
+[3]
+""", """\
+[1]
+[2]
+[3]
+[4]
+""", """\
+This
+is
+a
+list
+""",
+                                    ]
+
+        def program(self):
+            indices = range(4)
+
+            print(indices[0])
+            print(indices[1])
+            print(indices[2])
+            print(indices[3])
+
+    class indices_out_of_bounds(VerbatimStep):
+        """
+Now try `__program__` in the shell. What do you expect the output to be?
+        """
+
+        predicted_output_choices = ["0", "1", "2", "3", "4"]
+
+        correct_output = "Error"
+
+        expected_code_source = "shell"
+
+        program = "indices[4]"
+
+    class range_almost_the_same_as_list(VerbatimStep):
+        """
+`range(4)` is the same thing as `[0, 1, 2, 3]` ... almost. Try `__program__` in the shell.
+        """
+
+        expected_code_source = "shell"
+
+        program = "range(4)"
+
+    class range_versus_list(VerbatimStep):
+        """
+That's probably a bit surprising. If you're curious, the `0` represents the start of the range.
+`0` is the default start, so `range(4)` is equal to `range(0, 4)`.
+`4` is the end of the range, but the end is always excluded, so the last value is `3`.
+If you're confused now, don't worry about it.
+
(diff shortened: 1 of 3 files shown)


core/chapters/c06_lists.py
```diff
@@ -419,9 +419,140 @@ class index_error(Step):
...
-    class introducing_len_and_range(VerbatimStep):
+    class improve_with_list_and_loop(VerbatimStep):
         """
 There you go. `words[4]` and beyond don't exist, so trying that will give you an error.
+That first program is a bit repetitive. Let's improve it with a list and a loop!
+
+__program_indented__
+        """
+
+        def program(self):
```
alexmojaki, line 430:
```suggestion
        auto_translate_program = False

        def program(self):
```

That's a new part of the API. `['This', 'is', 'a', 'list']` isn't easily translated automatically, so the program needs to be translated manually.

core/chapters/c06_lists.py
```diff
@@ -419,9 +419,140 @@ class index_error(Step):
...
+            indices = [0, 1, 2, 3]
+
+            for index in indices:
+                print(index)
+                print(words[index])
+
+    class using_range_instead_of_indices(VerbatimStep):
+        """
+That's a bit better, but writing out `[0, 1, 2, ...]` isn't great, especially if it gets long.
+There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`.
```
alexmojaki, line 441:
```suggestion
There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`,
i.e. `indices = range(4)`.
```

Gotta be extra careful with these instructions so people don't get stuck.

core/chapters/c06_lists.py
```diff
@@ -419,9 +419,140 @@ class index_error(Step):
...
+It's not worth explaining that more right now.
+
+But you can easily convert it to a list: try `__program__` in the shell.
+
+What do you expect the output to be?
+        """
+        predicted_output_choices = [
+            "range(4)",
+            "range(0, 4)",
+            "list(range(4))",
```
alexmojaki, line 539:
```suggestion
            "list(range(4))",
            "list(range(0, 4))",
```

core/chapters/c06_lists.py
```diff
@@ -419,9 +419,140 @@ class index_error(Step):
...
+
+        program = "list(range(4))"
+
+    class introducing_len_and_range(VerbatimStep):
+        """
+That's just a demonstration to let you see a range in a more familiar form.
+You should almost never actually do that.
+
+If you're feeling overwhelmed, don't worry! All you need to know is that `range(n)`
+is very similar to the list `[0, 1, 2, ..., n - 2, n - 1]`.
```
alexmojaki, line 555:
```suggestion
is very similar to the list:

    [0, 1, 2, ..., n - 2, n - 1]
```

core/chapters/c06_lists.py
```diff
@@ -419,9 +419,140 @@ class index_error(Step):
...
+                print(words[index])
+
+    class printing_the_range(VerbatimStep):
+        """
+As you can see, the result is the same. Try this:
+
+    __copyable__
+    __program_indented__
+
+What do you expect the output to be?
```
alexmojaki, line 461:
```suggestion
```

alexmojaki: I just meant that the shell would say this

core/chapters/c06_lists.py
spamegg1: OK, didn't know that!

alexmojaki:
Thank you!

The steps after the new ones will need to be changed since they introduce `range()` again. While we're at it, let's introduce `len()` more slowly. How's this:

----

By the way, you can get the number of elements in a list (commonly called the length) using the `len()` function. Try len(words) in the shell.

(output predictions?)

----

Exercise: for any non-empty list `words`, print the last element. For example... (inputs and output)

(Now it's up to them to figure out `len(words) - 1`)

----

So in general, the valid indices are:

    [0, 1, 2, ..., len(words) - 2, len(words) - 1]

(Exercise to combine `range` and `len` (maybe only suggest specifically that in hints) to print each index and element of a list, like the program near the beginning but for any list)

----

Now it makes sense to continue as before with the exercise involving `to_find`. However I still think the whole thing is too fast and they could use some more practice with `range` and `len`. We could suggest a few exercises for them to practice on their own to get more comfortable:

- Print the numbers from 1 to n inclusive.
- Print some `word` n times.
- Print each word in `words` except for the last one.
- Print each word in `words` in reverse order.
- The bonus exercise from before where they put a separator between each word. Whether or not they did it then, it's easier now with `range/len`.

alexmojaki:
> OK, didn't know that!

Yeah this stuff was already so complicated before and the translation stuff is adding on to that so I've given up on documenting everything. I also think some potential contributors may have felt intimidated by that stuff so I'm discouraging newcomers from trying to do anything more than writing a draft in markdown.

spamegg1:
>Exercise: for any non-empty list words, print the last element. For example... (inputs and output)
(Now it's up to them to figure out len(words) - 1)

Should I give hints?

spamegg1:
Ah there is one more issue... After running the step with
```python
indices = range(4)

print(indices[0])
print(indices[1])
print(indices[2])
print(indices[3])
```
and then choosing the right predicted output, afterwards the console gets rid of `words`. Trying to access, say `words[2]` gives a "name `words` not defined" error. 

`python -m core.generate_static_files` also gives the same error if `words` is used in any `VerbatimStep` after the `indices` step. (`ExerciseStep`s are OK because `def solution(self, words: List[str]):` defines it.) 

Is this a bug? Or am I doing something wrong? I can get around it by using `len(['This', ...])` in the next step.

alexmojaki:
It's not a bug, the shell's locals are intentionally cleared when you run from the editor. Give them this instead of `len(words)` in the shell:

```python
    words = words = ['This', 'is', 'a', 'list']
    print(len(words))
```

Make it copyable and turn off auto-translation.

core/chapters/c06_lists.py
```diff
@@ -419,9 +419,143 @@ class index_error(Step):
...
+                print(words[index])
+
+    class using_range_instead_of_indices(VerbatimStep):
+        """
+That's a bit better, but writing out `[0, 1, 2, ...]` isn't great, especially if it gets long.
+There's a handy function `range` to do that part for you. Replace `[0, 1, 2, 3]` with `range(4)`,
+i.e. `indices = range(4)`.
+        """
+
+        program_in_text = False
```
alexmojaki, line 447:
```suggestion
        program_in_text = False
        auto_translate_program = False
```

spamegg1:
> It's not a bug, the shell's locals are intentionally cleared when you run from the editor. Give them this instead of `len(words)` in the shell:
> 
> ```python
>     words = words = ['This', 'is', 'a', 'list']
>     print(len(words))
> ```
> 
> Make it copyable and turn off auto-translation.

That's what I thought, already done! Will add copyable and no translation.

alexmojaki:
> > Exercise: for any non-empty list words, print the last element. For example... (inputs and output)
> > (Now it's up to them to figure out len(words) - 1)
> 
> Should I give hints?

Yes

alexmojaki:
Are you have trouble running the frontend test locally? You just need to download chromedriver and have the server running with `npm start`.

spamegg1:
> Are you have trouble running the frontend test locally? You just need to download chromedriver and have the server running with `npm start`.

Yes, they never worked for me, always some error... I think I'll close this for now and try later.

PR #271, opened 2022-01-10
Expand `range`

### LINKED ISSUE #261: Expanding explanation of range()

spamegg1:
fixes #261 
Let's try again!

To help you out I made a little summary:

### OLD:
01 - `introducing_subscripting`
02 - `index_error`
03 - `introducing_len_and_range`
04 - `range_len` (`VerbatimStep`, checked by `predict_output`)
05 - `index_exercise`
06 - `zip_exercise`
07 - `zip_longest_exercise`

### NEW:
01 - `introducing_subscripting`
02 - `index_error`
03 - `improve_with_list_and_loop`
04 - `using_range_instead_of_indices`
05 - `printing_the_range` **(checked by `predict_output`)**
06 - `indices_out_of_bounds`
07 - `range_almost_the_same_as_list`
08 - `range_versus_list`
09 - `introducing_len_and_range` **(modified, but same name)**
10 - `print_last_element`
11 - `range_len` **(now `ExerciseStep`, no longer checked by `predict_output`)**
12 - `index_exercise`
13 - `zip_exercise`
14 - `zip_longest_exercise`

I also added some comments in `test_frontend.py` to indicate at what step each thing is happening.


core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+            indices = range(4)
+
+            print(indices[0])
+            print(indices[1])
+            print(indices[2])
+            print(indices[3])
+
+    class indices_out_of_bounds(VerbatimStep):
+        """
+Now try `__program__` in the shell. What do you expect the output to be?
```
alexmojaki, line 506:
```suggestion
Now try `__program__` in the shell.
```

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+`0` is the default start, so `range(4)` is equal to `range(0, 4)`.
+`4` is the end of the range, but the end is always excluded, so the last value is `3`.
+If you're confused now, don't worry about it.
+
+There's a good reason for why `range(4)` is not actually a list - it makes programs faster and more efficient.
+It's not worth explaining that more right now.
+
+But you can easily convert it to a list: try `__program__` in the shell.
+
+What do you expect the output to be?
```
alexmojaki, line 538:
```suggestion
```

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
         def check(self):
             return "IndexError" in self.result
 
-    class introducing_len_and_range(VerbatimStep):
+    class improve_with_list_and_loop(VerbatimStep):
```
alexmojaki, line 423: I'd prefer to leave the names the same in earlier steps, so if someone was about to do `introducing_len_and_range` before then it won't skip content for them.

spamegg1: Hmm... OK

alexmojaki: specifically leave the names of the first 4 steps the same

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
 0
 1
 2
 3
 """, """\
-0
-This
 1
-is
 2
```
alexmojaki, line 475: Let's keep these predicted choices on `introducing_len_and_range` (currently `improve_with_list_and_loop`)

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+By the way, you can get the number of elements in a list (commonly called the *length*) using the `len` function.
+Try it by running this code:
+
+    __copyable__
+    __program_indented__
+        """
+
         auto_translate_program = False
 
+        predicted_output_choices = ["0", "1", "2", "3", "4"]
```
alexmojaki, line 573:
```suggestion
        predicted_output_choices = ["0", "1", "2", "3", "4", "5"]
```

core/chapters/c06_lists.py
spamegg1: adding them back...

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+So in general, the valid indices are:
+
+    [0, 1, 2, ..., len(words) - 2, len(words) - 1]
+
+Exercise: for any non-empty list `words`, print each index and element of the list. For example, if
 
+    __no_auto_translate__
+    words = ['This', 'is', 'a', 'list']
+
+your program should print:
```
alexmojaki, line 618:
```suggestion
No we can fix the program from earlier to work with any list. Fill in the `...`:

    __copyable__
    __no_auto_translate__
    words = ['This', 'is', 'a', 'list']

    for index in ...:
        print(index)
        print(words[index])

For the given example value of `words` it should print: 
```

core/chapters/c06_lists.py
spamegg1: Is this OK?
```python
        predicted_output_choices = ["""\
0
1
2
3
This
is
a
list
""",  """\
1
2
3
4
This
is
a
list
""", """\
This
is
a
list
0
1
2
3
""", """\
0
This
1
is
2
a
3
list
""", """\
1
This
2
is
3
a
4
list
""",                                     ]
```

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+a
+3
+list
+                    """),
+                ]
+
     class index_exercise(ExerciseStep):
         """
-Let's get some exercise! Given a list `things` and a value `to_find`,
+Using `range` and `len` in combination is a very important skill! At first, it may seem complicated.
```
alexmojaki, line 675:
```suggestion
```

core/chapters/c06_lists.py
spamegg1: done

core/chapters/c06_lists.py
spamegg1: done

core/chapters/c06_lists.py
spamegg1: added.

core/chapters/c06_lists.py
spamegg1: changed, I guess I should change the hints too?

spamegg1: any suggestions?

core/chapters/c06_lists.py
spamegg1: removed.

core/chapters/c06_lists.py
```diff
@@ -420,95 +420,277 @@ class index_error(Step):
...
+Print the word `Python` `n` times.
+
+Print each word in `words` except for the last one.
+
+Print each word in `words` in reverse order.
+
+Go back and revisit the bonus problem at the end of the [Introducing Lists page](#IntroducingLists),
+even if you've done it. It's now much easier with `range` and `len`!
+
+Let's get some more exercise! Given a list `things` and a value `to_find`,
```
alexmojaki, line 693:
```suggestion
If you're not quite comfortable with `range` and/or `len`,
practice and experiment with it for a bit.
Here are some simple exercises you can try on your own if you want.

For any non-empty list `words` and any positive number `n`:

- Print the numbers from `1` to `100` inclusive.
- Print your name 100 times.
- Print each word in a list `words` except for the last one.
- Print each word in `words` in reverse order, i.e. print the last word, then the second last word, etc.
- Revisit the bonus problem at the end of the [Introducing Lists page](#IntroducingLists),
whether or not you completed it. It's now much easier with `range` and `len`!

When you're ready, here's something a bit more challenging. Given a list `things` and a value `to_find`,
```

spamegg1: OK.

core/chapters/c06_lists.py
alexmojaki: No, use the same choices as before.

core/chapters/c06_lists.py
alexmojaki: yeah you should probably change them, no suggestions from me. i don't think they need to be perfect, they should solve this easily.

alexmojaki:
You know you can click "Commit suggestion" so you don't have to apply these manually?

spamegg1:
> You know you can click "Commit suggestion" so you don't have to apply these manually?

I do, I was afraid the test transcript would not get updated.

core/chapters/c06_lists.py
```diff
@@ -423,92 +423,319 @@ def check(self):
...
+list
+                    """),
+                ]
+
     class index_exercise(ExerciseStep):
         """
-Let's get some exercise! Given a list `things` and a value `to_find`,
+If you're not quite comfortable with `range` and/or `len`, practice and experiment with it for a bit.
+Here are some simple exercises you can try on your own if you want.
+For any non-empty list `words` and any positive number `n`:
```
alexmojaki, line 728:
```suggestion
```

tests/test_frontend.py
```diff
@@ -175,37 +175,42 @@ def _tests(driver):
...
-    predict_output(driver, editor, run_button, 0, 1)
+    predict_output(driver, editor, run_button, 1, 2)
 
     # Click OK
     driver.find_element_by_css_selector(".submit-prediction button").click()
 
-    # Course has moved on to next step
+    # Course has moved on to next step: indices_out_of_bounds
+    # Let's skip to the step: index_exercise
+    for _ in range(6):  # change this back to 6
```
alexmojaki, line 208: change this back to 6?

alexmojaki:
Looks great, thanks!

@anaisconce check it out!

core/chapters/c06_lists.py
```diff
@@ -423,92 +423,319 @@ def check(self):
...
+
+- Print the numbers from `1` to `100` inclusive.
```
alexmojaki, line 730: Hmmm, if you try any of these it shows a message because it's assuming you're doing the actual exercise. It'd be nice if this was the final text and the remaining steps were moved into a new page but that'll mess with existing user data.

core/chapters/c06_lists.py
```diff
@@ -423,92 +423,319 @@ def check(self):
...
+
+        def program(self):
+            words = ['This', 'is', 'a', 'list']
+            indices = [0, 1, 2, 3]
+
+            for index in indices:
+                print(index)
+                print(words[index])
+
+    class using_range_instead_of_indices(VerbatimStep):
```
alexmojaki, line 489:
```suggestion
    class range_len(VerbatimStep):
```

and rename the other `range_len`. I know the name doesn't make sense any more but it's fine.

core/chapters/c06_lists.py
spamegg1: I think moving them to the final text is a good idea. The challenging exercise will be a bit... challenging this way but it should be fine. `len` is not as confusing as `range`, which we eased them into.

tests/test_frontend.py
spamegg1: That's embarrassing :sweat_smile: I'll remove that.

core/chapters/c06_lists.py
spamegg1: what do you think?

core/chapters/c06_lists.py
spamegg1: OK

core/chapters/c06_lists.py
alexmojaki: You'll have to add some code to `book/store.js` to handle old user data in the new page structure. If they're on the step `index_exercise` then change it to `final_text`. If they're on a step after that, change the page slug to the new one.

core/chapters/c06_lists.py
```diff
@@ -494,21 +476,265 @@ class range_len(VerbatimStep):
...
+    1
+    is
+    2
+    a
+    3
+    list
+        """
+
+        hints = """
+Remember that earlier we used `range(4)` in place of `...`.
```
alexmojaki, line 683:
```suggestion
Remember that earlier we used `range(4)`.
```

PR #294, opened 2022-02-07
Add missing generate_po_file step to local test instructions

AndrewPluzhnikov:
Hello,

I just spent an hour trying to figure out why codes.json does not exist and why generate_static_files gave an assertion error.
I discovered that I needed to run generate_po_file first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,14 +136,16 @@ Several debuggers are provided, including [snoop](https://github.com/alexmojaki/
 
 1. Install Python 3.9 and [poetry](https://python-poetry.org/docs/#installation).
 2. Run `poetry install` to install Python dependencies.
-3. Run `poetry run python -m core.generate_static_files`. 
-4. In the `frontend` folder:
+3. Change into the `futurecoder` (top-level project) directory.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,14 +136,16 @@ Several debuggers are provided, including [snoop](https://github.com/alexmojaki/
 
 1. Install Python 3.9 and [poetry](https://python-poetry.org/docs/#installation).
 2. Run `poetry install` to install Python dependencies.
-3. Run `poetry run python -m core.generate_static_files`. 
-4. In the `frontend` folder:
+3. Change into the `futurecoder` (top-level project) directory.
```
alexmojaki, line 139: To me this step is implied, but if you think it's really necessary, it needs to come before `poetry install` since that installs the dependencies specified within the `futurecoder` directory.

alexmojaki:
Sorry about that! This translation stuff is very new and I'm still figuring out sensible workflows.

alexmojaki:
Updated the README to use the more convenient `generate.sh`. Thanks for the PR pushing me on this.

PR #312, opened 2022-02-26
Using `npx -y sass`

hangtwenty:
Smoketested it but mostly putting up as a memo


hangtwenty:
I tested it locally, but I guess slightly different behavior on CI (maybe using a different version of npm). Simple fix though, just a second...

alexmojaki:
`-o dist/course` isn't right. `course` corresponds to `frontend`, not `homepage`. The CSS files need to end up under `dist/static/css`.

hangtwenty:
I didn't notice `sass` in package.json, used `node-sass` just out of habit. (The intention of this PR was really just to use `npx -y`, it wasn't about which of `sass`/`node-sass` is used.)

alexmojaki:
Yeah it seems that `sass` is better than `node-sass`.

Why is it important to use `npx` for this?

In any case, I don't think package.json matters here since this happens outside of `frontend`. The `sass` command is installed by `npm install -g sass` just before in `full_build.sh`.

hangtwenty:
> Why is it important to use `npx` for this?

I try to avoid depending on `npm install -g`. (Much as we try to avoid Python installs that are not into a virtualenv. I use `asdf` on the npm/nodejs side.) 

With the change to `npx -y sass`, `build.sh` can be run without depending on the prereq.

alexmojaki:
OK, that makes sense. I haven't paid that much attention here because I don't expect many people to run build.sh. You might as well remove the `npm install -g`.

PR #313, opened 2022-02-28
WIP - making it an installable PWA (progressive web app)

hangtwenty:
⚠️ It doesn't quite work yet.

From these changes, locally, Lighthouse says it's an installable PWA. But it's not actually offline-ready from these changes. (So in my opinion it's not merge-able yet, but not as blocked as before.)


frontend/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "frontend",
   "version": "0.1.0",
-  "homepage": "/course/",
+  "homepage": "/",
```
hangtwenty, line 4: (Memo - would change back before merging. Just changed it for dev.)

frontend/public/index.html
```diff
@@ -15,6 +15,8 @@
     <meta property="og:type" content="website"/>
     <meta property="og:description" content="100% free and interactive Python course for beginners"/>
     <meta property="og:image" content="https://futurecoder.io/static/logo/bordered.png"/>
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
```
hangtwenty, line 18: Adding the manifest.json is the main thing that made the Lighthouse report say it's an installable PWA. But as mentioned in the PR description, it's not actually working yet.

frontend/src/serviceWorkerRegistration.js
```diff
@@ -0,0 +1,146 @@
+/* eslint-disable */
```
hangtwenty, line 1: This whole file is copied from https://github.com/cra-template/pwa/tree/main/packages/cra-template-pwa (which is linked from the CRA official docs re: PWA)

frontend/public/index.html
hangtwenty: image

hangtwenty: It "works" for a while, but then you see some `ERR_INTERNET_DISCONNECTED` errors in the logs, and after that point, upon refresh, it's just:

image

so, the caching or precaching seems to be messed up in the service worker ... But since I've followed the official CRA instructions and example, I'm not really sure why.

hangtwenty: Hm, at least as of [0912be0](https://github.com/alexmojaki/futurecoder/pull/313/commits/0912be0b8e26fa6c325824e3b9134f322d6a8d6a) (just pushed), it seems like it "works" until going offline. The bad part is that even when back online, it won't work again until all tabs have been closed.

hangtwenty: @alexmojaki Do you have any ideas based on the URLs in the error message?

alexmojaki: Those URLs are the first 4 things I see loading in my network tab in normal development. They're just the usual assets. So it sounds like nothing is loading at all. But it also looks like you're still using the dev server instead of building, maybe that messes things up?

hangtwenty: Sounds plausible that it's from something to do with using the dev server. Things weren't working for me when I built and served it though, on this branch... but it could have been some local mess.

hangtwenty:
cc: It'll probably be the weekend before I can get back around to this.

hangtwenty:
I'm not sure what else to try, but I'll think about it

frontend/craco.config.js
```diff
@@ -1,8 +1,17 @@
+const path = require('path');
+
 module.exports = {
   devServer: {
     headers: {
-      "Cross-Origin-Opener-Policy": "same-origin",
-      "Cross-Origin-Embedder-Policy": "require-corp"
+      'Cross-Origin-Opener-Policy'  : 'same-origin',
+      'Cross-Origin-Embedder-Policy': 'require-corp',
```
hangtwenty, line 7: (noise from my local code formatting preferences, can revert it)

build.sh
```diff
@@ -5,14 +5,22 @@ set -eux
...
+if [[ -z "${SKIP_TRANSLATIONS:-}" ]]; then
+    poetry run python -m translations.generate_po_file
+    poetry run python -m core.generate_static_files
+else
+    # If SKIP_TRANSLATIONS is set, set a default language for this build
+    if [[ -z "${FUTURECODER_LANGUAGE:-}" ]]; then
+        export FUTURECODER_LANGUAGE="en"
+    fi
+fi
+
```
alexmojaki, line 17: Now that we're outputting the build to `course`, I don't see a need for this change.

alexmojaki:
package-lock.json has more than double in size, that's a bit scary

frontend/src/RunCode.js
```diff
@@ -31,9 +31,10 @@ function initWorker() {
 initWorker();
 
 const channel = makeChannel({serviceWorker: {scope: "/course/"}});
-if (channel?.type === "serviceWorker") {
-  navigator.serviceWorker.register("./service-worker.js");
-}
+// // "also you should probably comment out this code in RunCode.js just to avoid having multiple places that register the service worker:"
+// if (channel?.type === "serviceWorker") {
+//   navigator.serviceWorker.register("./service-worker.js");
+// }
```
alexmojaki, line 37: You can remove this. Do you think you can make the service worker always register, but by default disable the caching on localhost from within the service worker itself?

frontend/src/index.js
```diff
@@ -4,9 +4,19 @@ import {App} from './App';
...
     <Provider store={store}>
       <App/>
     </Provider>,
   document.getElementById("root")
 );
+
+
+// If you want your app to work offline and load faster, you can change
+// unregister() to register() below. Note this comes with some pitfalls.
+// Learn more about service workers: https://cra.link/PWA
```
alexmojaki, line 21: Remove

frontend/src/service-worker.js
```diff
@@ -1,15 +1,72 @@
...
+// service worker, and the Workbox build step will be skipped.
 
-const fetchListener = serviceWorkerFetchListener();
+import { clientsClaim } from 'workbox-core';
+import { ExpirationPlugin } from 'workbox-expiration';
+import { precacheAndRoute, createHandlerBoundToURL } from 'workbox-precaching';
+import { registerRoute } from 'workbox-routing';
+import { StaleWhileRevalidate } from 'workbox-strategies';
 
-addEventListener('fetch', fetchListener);
```
alexmojaki, line 7: Keep this, sync-message is still important

frontend/src/service-worker.js
```diff
@@ -1,15 +1,72 @@
...
+    return true;
+  },
+  createHandlerBoundToURL(process.env.PUBLIC_URL + '/index.html')
+);
+
+// An example runtime caching route for requests that aren't handled by the
+// precache, in this case same-origin .png requests like those from in public/
+registerRoute(
+  // Add in any other file extensions or routing criteria as needed.
+  ({ url }) => true, // XXX Caching everything for now. It breaks firebase, but it makes most of the PWA work offline. We can narrow it down later.
```
alexmojaki, line 55: Needs to be narrowed down now. Bootstrap should probably be moved into futurecoder rather than served from a CDN. Then that just leaves pyodide.

I'd still like to understand:

1. Why caching from other origins works at all, and if it can be relied upon.
2. Whether it would be good to have precaching for the pyodide stuff anyway

frontend/src/serviceWorkerRegistration.js
```diff
@@ -0,0 +1,145 @@
...
+// Otherwise webpack can fail silently
+// https://github.com/facebook/create-react-app/issues/8014
```
alexmojaki, line 3: I don't think this should be needed here

frontend/src/service-worker.js
hangtwenty: Good catch — I replaced `service-worker.js` wholesale with the one from `cra-template-pwa`. Will put that back.

frontend/src/service-worker.js
```diff
@@ -27,39 +27,62 @@ precacheAndRoute(self.__WB_MANIFEST);
...
+    ({url}) => {
+      const urlString = url.toString();
+      if (urlString.startsWith('https://cdn.jsdelivr.net/')) {
+        return true;
+      } else if (urlString.startsWith('https://cdnjs.cloudflare.com')) {
+        return true;
+      } else if (urlString.startsWith('https://stackpath.bootstrapcdn.com/bootstrap/')) {
+        // TODO: move bootstrap inline rather than getting from CDN
+        return true;
+      } else if (urlString.startsWith('https://futurecoder.io')) {
```
alexmojaki, line 62: Rather just ensure it contains `futurecoder.io`, to account for translations like https://es-latam.futurecoder.io/course/

frontend/src/service-worker.js
```diff
@@ -27,39 +27,62 @@ precacheAndRoute(self.__WB_MANIFEST);
...
-    ],
-  })
-);
+if (process.env.NODE_ENV === "production" || process.env.NODE_NEV === "dev-pwa") {
+  registerRoute(
+    ({url}) => {
+      const urlString = url.toString();
+      if (urlString.startsWith('https://cdn.jsdelivr.net/')) {
+        return true;
+      } else if (urlString.startsWith('https://cdnjs.cloudflare.com')) {
```
alexmojaki, line 57: Please combine all these cases with `||`

frontend/src/RunCode.js
hangtwenty: Good thinking. I think it's working this way now.

hangtwenty: (Feel free to resolve when you've tested it too)

build.sh
hangtwenty: Alrighty, changed it so that the `npm run pwa` command no longer exercises `build.sh`, and reverted this part of `build.sh`

frontend/src/service-worker.js
hangtwenty: OK, now there's more specific rules in `service-worker.js`, and they're only applied when it's a build (i.e. `NODE_ENV=production`).

However, when testing offline, it errors out on the auth/firebase stuff, as we were seeing before. So, offline-readiness might be blocked until changes there.

frontend/src/service-worker.js
```diff
@@ -48,25 +50,20 @@ registerRoute(
...
-      } else if (urlString.startsWith('https://futurecoder.io')) {
-        return true;
-      } else if (urlString.startsWith('https://fonts.googleapis.com')) {
-        return true;
-      } else if (urlString.startsWith('https://pyodide-cdn2.iodide.io')) {
-        return true;
+      if (
+        urlString.startsWith('https://cdn.jsdelivr.net/')
+        || urlString.startsWith('https://cdnjs.cloudflare.com')
+        || urlString.startsWith('https://pyodide-cdn2.iodide.io')
```
alexmojaki, line 60: Could you add comments for when these are used? What happened to fonts.googleapis.com?

hangtwenty:
> package-lock.json has more than double in size, that's a bit scary

I think it's because I am using a newer version of npm and it auto-upgrades the format.  
The one in this branch has `"lockfileVersion": 2`, and in master it has   `"lockfileVersion": 1`. npm does the upgrade without asking, which makes it seem harmless, but maybe not. Want me to revert it / use older npm?

frontend/src/service-worker.js
hangtwenty: @alexmojaki I realized that one was only used in `homepage` not within `frontend`, so it seems unnecessary. But I can put it back.

Adding comments 👍

frontend/src/service-worker.js
```diff
@@ -1,15 +1,91 @@
...
+
+if (process.env.NODE_ENV === 'production') {
+  registerRoute(
+    ({url}) => {
+      const urlString = url.toString();
+      if (
+        urlString.startsWith('https://cdn.jsdelivr.net/') // Pyodide
+        || urlString.startsWith('https://pyodide-cdn2.iodide.io') // Only used when we are testing bleeding-edge pyodide
+        || urlString.startsWith('https://cdnjs.cloudflare.com') // Only used by homepage, so probably not needed here
+        || urlString.startsWith('https://fonts.googleapis.com') // Only used by homepage, so probably not needed here
```
hangtwenty, line 61: @alexmojaki what do you think, shall I remove these?

```
        || urlString.startsWith('https://cdnjs.cloudflare.com') // Only used by homepage, so probably not needed here
        || urlString.startsWith('https://fonts.googleapis.com') // Only used by homepage, so probably not needed here
```

alexmojaki:
I'm sure I need to upgrade my npm at some point, but I'd like that change to be separate from adding dependencies.

Surely you still need to update package-lock.json since you added stuff to package.json?

frontend/src/service-worker.js
alexmojaki: Yeah, remove them.

hangtwenty:
@alexmojaki Re: package-lock.json, just looked into this more — FutureCoder's README currently says to use node.js 14.15.4 and npm 7.5.2. So I ensured I was using those versions and did an npm install, to get the updated package-lock.json. It's still lockfileVersion 2, actually. Checked the docs and that's expected for npm 7: https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion

So it seems that I'd need to use npm 6.x to generate the package-lock.json in a way that doesn't convert to lockfileVersion 2. But that would diverge from the README.

Should we just let it upgrade?

hangtwenty:
Oh, you know what? I can just remove the explicit dependencies on `workbox-*` that I'd added. Just tested locally to confirm, they're not actually needed. They're dependencies of CRA (or more specifically the Workbox Webpack plugin).

So, I'm just removing those added dependencies, and then package-lock.json can stay untouched for now.

alexmojaki:
Yes, I've been using npm v6 for a while despite the README. I don't know how this happened, but it's good to be pushed to fix this mess. And yes, I should probably use `asdf` or add a `.nvmrc` file to the repo. I don't think the exact version of node still matters, but it's good for everyone to be on the same page.

build.sh
```diff
@@ -9,10 +9,10 @@ poetry run python -m translations.generate_po_file
 poetry run python -m core.generate_static_files
 
 cd frontend
-REACT_APP_LANGUAGE=$FUTURECODER_LANGUAGE CI=false npm run build
+NODE_ENV=production REACT_APP_LANGUAGE=$FUTURECODER_LANGUAGE CI=false npm run build
```
alexmojaki, line 12: > There is also a built-in environment variable called NODE_ENV. You can read it from process.env.NODE_ENV. When you run npm start, it is always equal to 'development', when you run npm test it is always equal to 'test', and **when you run npm run build to make a production bundle, it is always equal to 'production'. You cannot override NODE_ENV manually.** This prevents developers from accidentally deploying a slow development build to production.

https://create-react-app.dev/docs/adding-custom-environment-variables/

frontend/src/RunCode.js
```diff
@@ -202,6 +200,7 @@ export const _runCode = wrapAsync(async function runCode({code, source}) {
   });
 
   if (error) {
+    console.warn(error)
```
alexmojaki, line 203: What made you add these? I think this one should be removed because `error` is a very large object at this stage and logging it might redundantly send it to Sentry as a breadcrumb. Not sure if the other one is useful in dev.

frontend/src/service-worker.js
```diff
@@ -1,15 +1,89 @@
...
+    }),
+  );
+}
 
-addEventListener('activate', function (e) {
-  e.waitUntil(self.clients.claim());
+// This allows the web app to trigger skipWaiting via
+// registration.waiting.postMessage({type: 'SKIP_WAITING'})
+self.addEventListener('message', (event) => {
+  if (event.data && event.data.type === 'SKIP_WAITING') {
```
alexmojaki, line 86: We're not sending any such message anywhere. Just replace this with `skipWaiting` from `workbox-core`.

frontend/public/index.html
```diff
@@ -3,8 +3,7 @@
 <head>
     <meta charset="utf-8"/>
     <link rel="shortcut icon" type="image/png" href="/static/logo/favicon_32px.png"/>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-          integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
+    <link rel="stylesheet" href="%PUBLIC_URL%/bootstrap-4.4.1.min.css">
```
alexmojaki, line 6: Does webpack/workbox now pick this up as something to precache?

Also is this the best way to include bootstrap? I feel like there's supposed to be some better way where you `npm install` and then `import` something and it bundles everything just right.

frontend/src/service-worker.js
```diff
@@ -1,15 +1,89 @@
...
+  // even if you decide not to use precaching. See https://cra.link/PWA
+  precacheAndRoute(self.__WB_MANIFEST);
+}
+
+// Set up App Shell-style routing, so that all navigation requests
+// are fulfilled with your index.html shell. Learn more at
+// https://developers.google.com/web/fundamentals/architecture/app-shell
+const fileExtensionRegexp = new RegExp('/[^/?]+\\.[^/]+$');
+registerRoute(
+  // Return false to exempt requests from being fulfilled by index.html.
```
alexmojaki, line 31: I don't understand what this is doing, do you?

frontend/src/service-worker.js
```diff
@@ -1,15 +1,89 @@
...
+      return false;
+    }
+
+    // Return true to signal that we want to use the handler.
+    return true;
+  },
+  createHandlerBoundToURL(process.env.PUBLIC_URL + '/index.html'),
+);
+
+if (process.env.NODE_ENV === 'production') {
```
alexmojaki, line 53: Please combine the two uses of this.

frontend/src/service-worker.js
```diff
@@ -1,15 +1,89 @@
...
+    new StaleWhileRevalidate({
+      cacheName: 'everything',
+      cacheExpiration: {
+        maxEntries: 50,
+      },
+      cacheableResponse: {statuses: [0, 200]},
+      plugins: [
+        // Ensure that once this runtime cache reaches a maximum size the
+        // least-recently used stuff is removed.
+        new ExpirationPlugin({maxEntries: 30}),
```
alexmojaki, line 77: 1. Why are there two different `maxEntries`?
2. What does a status of `0` mean?

frontend/src/RunCode.js
hangtwenty: I added them when I was trying to nail down inconsistencies, but no problem to remove them. Will do

frontend/src/service-worker.js
hangtwenty: So just replacing with 

```javascript
self.skipWaiting()
```

right?

(What was there is just what was from the `cra-template-pwa` example)

frontend/public/index.html
hangtwenty: It's probably not the best way. I didn't look into the Webpack way, but I could

frontend/src/service-worker.js
alexmojaki: > self.skipWaiting()

No, import skipWaiting from workbox-core, like clientsClaim.

> (What was there is just what was from the cra-template-pwa example)

I know, just cleaning up, and I do want to call skipWaiting eagerly for the non-caching stuff.

frontend/public/index.html
alexmojaki: Let's just keep this in mind for now.

frontend/src/service-worker.js
alexmojaki: I looked at the comments and the link but still wasn't quite sure I understood the code.

frontend/src/service-worker.js
hangtwenty: 1 - oops :)
2 - Including `0` here has something to do with support for CORS, but I'm unsure whether we need it. It's another thing I took from official boilerplate/examples, but in this case from Workbox's docs:

image

That screenshot is from [here](
https://developer.chrome.com/docs/workbox/modules/workbox-cacheable-response/#why-are-there-different-defaults). More from it:

image

In practice it seems like people use it "by default" if there are third-party resources, but Workbox's cacheableResponse defaults to only `[200]` because it would be astonishing if `0` was invisibly part of the defaults.

I found [this post][(https://whatwebcando.today/articles/opaque-responses-service-worker/) to be clarifying:

> **Option 2: Ignoring opaque responses**
> Is it then better to ignore the opaque responses? This is what Workbox does by default and this is what definitely makes sense for cache-first strategies. **What we lose in this case is the ability to serve any result of the cross-origin request in offline mode, which for some apps might be a major downside.**

(**Emphasis** mine.) I haven't done an A/B test without 0 though.

alexmojaki: Thanks for looking into it. 0 sounds better, just add a little comment and include a link or two.

frontend/src/service-worker.js
hangtwenty: Tested and it seems to work the same if it's just `self.skipWaiting()` eagerly, so we'll go with that.

frontend/src/service-worker.js
hangtwenty: OK cool, looks like it can all go inside of `"production"`. Tested both `npm run start` (to make sure there were no regressions) and `npm run pwa`.

frontend/src/service-worker.js
hangtwenty: > OK cool, looks like it can all go inside of `"production"`. Tested both `npm run start` (to make sure there were no regressions) and `npm run pwa`.

frontend/src/service-worker.js
alexmojaki: No, use the workbox function or copy the source code so that it's inside the install/activate event listener. That's what's always suggested and I'd like to stick to that unless I know why it's usually like that and why it wouldn't apply to us. That's still sufficiently eager.

frontend/src/service-worker.js
hangtwenty: I tested it, and it _seems_ like the same behavior for now, if we remove the `0` argument... but since "offline" currently errors out due to the firebase stuff, we don't have a base-case to compare it to and make sure it works. I think it would be easy to forget this, so I suggest we keep `0` for now, and then after things are working-offline without firebase errors, we could check again to see if `0` is unnecessary.

frontend/src/service-worker.js
```diff
@@ -1,15 +1,79 @@
...
+      // ^ Q: What's status 0 mean?
+      // ^ A: It's an opaque response, and it relates to third-party resources (and CORS).
+      // https://github.com/alexmojaki/futurecoder/pull/313/files/f5f0fded9f44fefd0bf44b99ee45f54a8badc272#r829408930
+      plugins: [
+        new ExpirationPlugin({maxEntries: 30}),
+      ],
+    }),
+  );
+}
+
```
hangtwenty, line 78: @alexmojaki service-worker.js is relatively clean now. Good suggestions about cleaning up stuff from the boilerplate!

frontend/src/service-worker.js
alexmojaki: Since we've whitelisted the domains, firebase shouldn't be relevant here. I'm happy to leave the 0.

alexmojaki:
I deployed a preview: https://futurecoder-io--pwa-tfvdlu0q.web.app/course/

It works on the first load, and after refreshing once, but after refreshing a second time it can no longer load pyodide. It looks cached, so I don't know what's wrong.

Screenshot from 2022-03-18 11-03-32

However I can only reproduce this in non-incognito Chrome, on the deployed preview. I don't have this problem:

- With localhost
- In incognito mode
- In Firefox

I've cleared the site data over and over, every time it breaks after two refreshes.

Another problem is birdseye. The way that birdseye works is to open a new tab at `/course/birdseye/`. But with the new service worker it just shows the same page as `/course/`. I'm guessing this is the "App Shell-style routing". Maybe we should add another case to return `false`, but I'm also wondering if we should just remove that route entirely. How is it helping?

I also think we should probably be precaching the entire `birdseye` folder. If someone installs the PWA and goes offline, birdseye should work even if they'd never used it while online. That's part of what precaching is for, right? Pity we can't do the same for Python Tutor.

alexmojaki:
By the way, since you're now a collaborator, you can make PRs using branches instead of forks. Doing that will give GHA access to the secrets that allow automatically deploying a firebase preview every time you push to the branch. I should have mentioned this earlier but I only just remembered the branch/fork part and was wondering why you being a collaborator wasn't making it work already.

hangtwenty:
> I deployed a preview: [futurecoder-io--pwa-tfvdlu0q.web.app/course](https://futurecoder-io--pwa-tfvdlu0q.web.app/course/)

Thanks for doing that.

> It works on the first load, and after refreshing once, but after refreshing a second time it can no longer load pyodide. It looks cached, so I don't know what's wrong.
> 
> Screenshot from 2022-03-18 11-03-32
> 
> However I can only reproduce this in non-incognito Chrome, on the deployed preview. I don't have this problem:
> 
> * With localhost
> * In incognito mode
> * In Firefox
> 
> I've cleared the site data over and over, every time it breaks after two refreshes.

That's maddening!

hangtwenty:
> I also think we should probably be precaching the entire `birdseye` folder. If someone installs the PWA and goes offline, birdseye should work even if they'd never used it while online. That's part of what precaching is for, right? Pity we can't do the same for Python Tutor.

Good idea. Maybe birdseye could be a subsequent PR, since it's an enhancement of sorts?

I think the firebase issue is still a blocker to merging though, because it breaks loading (not just firebase).

hangtwenty:
I've switched my `origin` to `alexmojaki/futurecoder` and I'll work off the branch there now. Shall I open a new PR and close this one, to keep things straight?

Issue #163, opened 2021-07-06
Allow using from China


alexmojaki:
Original report from Slack:

> hello, I am from China, I am very happy to find this platform, and I want to use it in my course, but my students do not connect the google, I want to know there are some others ways for us to use this platform. think you!

Firebase is blocked in China, need to find a workaround.

alexmojaki:
Looks like several potential users are running into this and it's becoming more important, especially since someone has started translating futurecoder to Chinese.

For starters, there should at least be an error message when the site fails to connect to firebase. Right now it just gets stuck on loading.

Options to actually make the site work:

1. Set up a server to proxy requests to firebase.
2. Use local storage as a fallback, which might also help in case of other connection problems, but wouldn't allow logging into the account on another computer.
3. Find an alternative to firebase with the same features and reimplement all that code. Auth is the biggest concern here.

hangtwenty:
I think that https://supabase.com might be worth considering. (I've heard good things, anyways!)

alexmojaki:
That's what I thought, but when I looked more deeply it actually seemed very difficult. I'd be thrilled if someone could manage, because it seems like in the long term I'm going to have to leave firebase over this.

hangtwenty:
I've heard that Supabase can be a good alternative to Firebase, so I'm curious if Supabase would be accessible to the users who are currently unable to access Firebase. I haven't tried that migration myself [but I found this](https://github.com/supabase-community/firebase-to-supabase)

alexmojaki:
Interesting! I'm not in a hurry to migrate, but it'd be nice if this was production ready: https://github.com/supabase/ui

berhalak:
Hi Alex.

Some insights that might help. I have two accounts one works and one doesn't. For the one that doesn't the error is here:

![image](https://user-images.githubusercontent.com/11277225/157984359-df654770-cfb2-4821-ab56-54ec017b8053.png)

For the account that doesn't work currentUser.uid is my email, not uid. When I'm using the second account, this uid is some valid alphanumeric value.

alexmojaki:
Thanks for looking into this @berhalak. Looks like I messed up the migration of users from django to firebase. Probably not worth fixing now.

PR #317, opened 2022-03-13
[experimental] Wrap firebase/googleapis calls with try/catch

hangtwenty:
While I was testing, something caused an infinite redirect to happen, which caused a lot of calls to Firebase, then Firebase temporarily blocked my browser. So, I figured I'd try wrapping the Firebase stuff in try/catch; it might help with issue #163.

With these changes, the course works (and code runs) without access to `firebaseio.com` — tested by blocking network requests to that domain. It still breaks once I block `googleapis.com`, **but,** it's not erroring-out. It's just staying on "Loading..." because `isLoaded` depends on `state.user.uid`. I'm guessing that there's a way to refactor such that `isLoaded` can know it's loaded without depending on `state.user.uid` — or perhaps dummy info could be put into `state.user` when auth stuff fails.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -438,7 +438,11 @@ const MenuPopup = ({user}) =>
             onClick={() => {
               close();
               bookSetState("user.uid", null)
-              firebase.auth().signOut();
+              try {
+                firebase.auth().signOut().catch(console.warn);
+              } catch (err) {
+                console.warn(err)
+              }
             }}
           >
             <FontAwesomeIcon icon={faSignOutAlt}/> {terms.sign_out}
--- a/frontend/src/book/store.js
+++ b/frontend/src/book/store.js
@@ -34,24 +34,37 @@ const firebaseConfig = {
   measurementId: "G-ZKCE9KY52F",
 };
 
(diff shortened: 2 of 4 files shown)


frontend/src/App.js
```diff
@@ -438,7 +438,11 @@ const MenuPopup = ({user}) =>
             onClick={() => {
               close();
               bookSetState("user.uid", null)
-              firebase.auth().signOut();
+              try {
+                firebase.auth().signOut().catch(console.warn);
+              } catch (err) {
+                console.warn(err)
```
hangtwenty, line 444: May seem redundant, but the the `.catch()` is to handle the `.signOut()` promise, whereas the `try`/`catch` is to handle an error when invoking `firebase.auth()`.

(Same is true of a couple of other places in this changeset.)

alexmojaki:
Whenever you run code, it checks if that code is correct for whatever step you're on. So it needs to know what step you're on, and the way to get that data initially is to load the user, particularly so that existing users come back to the step they were on last time. Previously you could get errors if you ran code early enough, before the user had loaded.

This comes back to #269 and the stuff we chatted about. Let's leave firebase alone for now. I also don't love the idea of wrapping everything in try/catch.

PR #320, opened 2022-03-23
Installable PWA

hangtwenty:
ℹ️  _This branch adds configuration to the service worker to make it a PWA, in order to make it installable and work without reliable internet access._

⏮️  This picks up from my previous PR https://github.com/alexmojaki/futurecoder/pull/313 which was from a fork, but now it's a branch in this repository so that we can run the workflow for test deploys.

----

⚠️ We still may not want to merge it before making changes to the auth/firebase stuff. As it stands, when it's offline, there are errors in the dependencies on firebase auth, and ultimately the errors keep it stuck on "Loading...". (We agreed when we talked about https://github.com/alexmojaki/futurecoder/pull/317 that we don't want to put try/catch around Firebase stuff everywhere, as there needs to be a way to keep/restore user progress even without access to firebase.)


hangtwenty:
One of the open issues, from the previous PR:

> I deployed a preview: [futurecoder-io--pwa-tfvdlu0q.web.app/course](https://futurecoder-io--pwa-tfvdlu0q.web.app/course/)
> 
> It works on the first load, and after refreshing once, but after refreshing a second time it can no longer load pyodide. It looks cached, so I don't know what's wrong.
> 
> Screenshot from 2022-03-18 11-03-32
> 
> However I can only reproduce this in non-incognito Chrome, on the deployed preview. I don't have this problem:
> 
> * With localhost
> * In incognito mode
> * In Firefox
> 
> I've cleared the site data over and over, every time it breaks after two refreshes.
> 
> Another problem is birdseye. The way that birdseye works is to open a new tab at `/course/birdseye/`. But with the new service worker it just shows the same page as `/course/`. I'm guessing this is the "App Shell-style routing". Maybe we should add another case to return `false`, but I'm also wondering if we should just remove that route entirely. How is it helping?
> 
> I also think we should probably be precaching the entire `birdseye` folder. If someone installs the PWA and goes offline, birdseye should work even if they'd never used it while online. That's part of what precaching is for, right? Pity we can't do the same for Python Tutor.

alexmojaki:
I'm still getting the double-refresh bug, and it's also happening in incognito, but not Firefox. Just to confirm, do you see this problem too?

hangtwenty:
@alexmojaki I don't think I've reproduced the double-refresh bug, exactly, but I see this with the preview deploy even on first load:

```
The FetchEvent for "https://cdn.jsdelivr.net/pyodide/v0.19.0/full/pyodide.js" resulted in a network error response: the promise was rejected.
```

and

```
The FetchEvent for "https://cdn.jsdelivr.net/pyodide/v0.19.0/full/pyodide.js" resulted in a network error response: the promise was rejected.
service-worker.js:1 Uncaught (in promise) TypeError: Type error
Worker.js:24 Uncaught (in promise) DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'https://cdn.jsdelivr.net/pyodide/v0.19.0/full/pyodide.js' failed to load.
    at A (https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:15509)
    at https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:16457
    at Module.<anonymous> (https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:16669)
    at r (https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:110)
    at https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:912
    at https://futurecoder-io--refs-pull-320-merge-htu63cug.web.app/course/static/js/Worker.f8349279.worker.js:1:921
```

Though oddly, Pyodide is actually working despite those errors. At first. Then double-refresh leads to an error.

And yeah, seeing it only in Chrome, with the preview deploy. I don't see the issue in Firefox, even with the preview deploy.

So weird.

alexmojaki:
I think now is the time to investigate more deeply what happens when service workers try to intercept requests to other origins.

alexmojaki:
I just completed a big refactoring in #321 so I've merged master into here. The extra service worker registration is now in TaskClient.js and needs to be removed there again in this PR.

hangtwenty:
Welp, adding CacheableResponsePlugin explicitly was worth a shot, but it's still exhibiting the same behavior.

alexmojaki:
I took a shot in the dark...and apparently it worked?!?! Please confirm that the problem looks fixed on your end.

hangtwenty:
@alexmojaki Hm, I've been getting a new error since [bc7df16](https://github.com/alexmojaki/futurecoder/pull/320/commits/bc7df16c6ad034669aa783ef7e9c2e760a3e556e) —

```
  File "/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'core.init_pyodide'
```

Or with more context:

```
pyodide.asm.js:14 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'core.init_pyodide'

    at new_error (pyodide.asm.js:14:247183)
    at pyodide.asm.wasm:0xe6eb0
    at pyodide.asm.wasm:0xe6fae
    at Object.Module._pythonexc2js (pyodide.asm.js:14:944230)
    at Object.Module.callPyObjectKwargs (pyproxy.gen.js:391:12)
    at Object.Module.callPyObject (pyproxy.gen.js:397:17)
    at PyProxyClass.apply (pyproxy.gen.js:1185:19)
    at Object.apply (pyproxy.gen.js:1062:18)
    at Object.pyimport (api.js:313:27)
    at Worker.js:11:11
```

Tested in multiple browsers, same error in all of them.

alexmojaki:
I'm guessing this only happens locally and not in the preview? You need to `poetry install` and `generate_static_files` again, your `python_core.tar` is stale.

hangtwenty:
In the preview, I'm still seeing firebase errors when I go offline (and it still stays stuck on "Loading...")

```
trycatch.ts:182          POST https://www.googleapis.com/identitytoolkit/v3/relyingparty/getAccountInfo?key=[...] net::ERR_INTERNET_DISCONNECTED
[...]
17:02:08.452 trycatch.ts:182          GET https://futurecoder-io-default-rtdb.firebaseio.com/users/pMbI5MYEQBhWOHdjfkAjQig2lcc2.json?auth=[...] net::ERR_INTERNET_DISCONNECTED
[...]
17:02:08.996 load-pyodide.js:213 Loading distutils
17:02:08.996 load-pyodide.js:239 Loading distutils from https://cdn.jsdelivr.net/pyodide/v0.19.0/full/distutils.js
17:02:09.029 load-pyodide.js:273 Loaded distutils
17:02:09.031 pyodide.asm.js:14 Python initialization complete
17:02:10.456 trycatch.ts:182          GET https://futurecoder-io-default-rtdb.firebaseio.com/users/pMbI5MYEQBhWOHdjfkAjQig2lcc2.json?auth=[...] net::ERR_INTERNET_DISCONNECTED
[...]
17:02:14.524 trycatch.ts:182          GET https://futurecoder-io-default-rtdb.firebaseio.com/users/pMbI5MYEQBhWOHdjfkAjQig2lcc2.json?auth=[...] net::ERR_INTERNET_DISCONNECTED
```

alexmojaki:
That's completely expected, nothing we've done would allow firebase to work offline. But we should address that now, as removing cross origin isolation comes with a cost that needs to be justified.

First, remove the cross origin headers from the dev server in the craco config.

Second, replace the manual REST API calls to firebase with firebase SDK calls, something like `firebase.database().ref(pathString).set(value)`. The goal is for firebase to automatically fallback to local offline storage. Leave the stuff that posts to `code_entries` unchanged so that it doesn't accumulate an unlimited number of those while offline.

Finally, remember that birdseye is still completely broken in this PR, offline or not, and that has to be fixed. Precaching birdseye would also be nice but is less essential.

hangtwenty:
Oh, when you said —

> I took a shot in the dark...and apparently it worked?!?! Please confirm that the problem looks fixed on your end.

— I thought "it" was offline. I didn't know why it was expected to work offline just from that recent change, but now I understand you were talking about something more specific.

I'm not prepared to dig into the firebase stuff to make it offline-friendly, at least at this time, because I am not using firebase on other projects so it would be a bit of a rabbithole for me.

hangtwenty:
Re: birdseye, I _think_ we'd need to get Workbox to [pre]cache everything in `frontend/public`. Pretty sure that can be accomplished by configuring [Workbox's `InjectManifest` options, i.e. `include`](https://developer.chrome.com/docs/workbox/reference/workbox-build/#:~:text=One%20or%20more%20specifiers%20used%20to%20include%20assets%20in%20the%20precache%20manifest.). Currently those options are hidden inside craco's CRA config, and they can be overridden but I have not gotten it working yet. I'll put it on its own branch, because I don't want to push up the current branch in a broken state.

hangtwenty:
Re: birdseye working on this branch (online), I think it's now fixed. Tested locally and on the preview deploy. LMK if you think it is fixed too.

hangtwenty:
The fix in [ab45304](https://github.com/alexmojaki/futurecoder/pull/320/commits/ab45304d5559df0dee86b8304016cf5b1799a306) doesn't result in precaching birdseye, to be sure. I don't think it prevents it from being solved, either. The change in [ab45304](https://github.com/alexmojaki/futurecoder/pull/320/commits/ab45304d5559df0dee86b8304016cf5b1799a306) is about that `registerRoute` call; we'll get precaching of birdseye to work as part of `precacheAndRoute(self.__WB_MANIFEST)` (but currently blocked, see #325).

frontend/craco.config.js
```diff
@@ -1,17 +1,19 @@
...
-  },
   // Output to ./course (instead of ./build)
   webpack: {
     configure: (webpackConfig, {env, paths}) => {
       paths.appBuild = webpackConfig.output.path = path.resolve('course');
       return webpackConfig;
     },
   },
+
+  // Staged for deletion -- the CORS headers override should not be needed as of recent commits in https://github.com/alexmojaki/futurecoder/pull/320
```
alexmojaki, line 12: just delete them

alexmojaki:
> but now I understand you were talking about something more specific.

lol yes I was talking about the double refresh bug that we'd been struggling with for a while.

Looks like birdseye is fixed. But I still think we'd probably be better off without that app shell routing at all to prevent similar problems in the future. futurecoder uses `#` fragments for client-side routing, not paths.

hangtwenty:
@alexmojaki Cool, I've removed app-shell routing and tested locally again, and it seems like nothing broke. (When I tried removing it before, I thought it broke something. So either that problem's gone, or there was not even a problem.)

alexmojaki:
Thanks. Keep looking into precaching birdseye, I'll try dealing with firebase when I have time.

frontend/src/service-worker.js
```diff
@@ -1,15 +1,54 @@
...
 
-addEventListener('install', function (e) {
-  e.waitUntil(self.skipWaiting());
-});
+clientsClaim();
 
-addEventListener('activate', function (e) {
-  e.waitUntil(self.clients.claim());
-});
+if (process.env.NODE_ENV === 'production') {
```
alexmojaki, line 20: I don't think this actually helps. The dev server doesn't serve service-worker.js, right? So we always essentially `npm run build` to use the service worker, which implies `NODE_ENV === 'production'` automatically. Even if service-worker.js is copied into `public` and the dev server is used, service-worker.js won't know that.

alexmojaki: I think we need a `REACT_APP_` env var so that we can control it, and turning it on would enable caching futurecoder-related things (i.e. everything other than pyodide) whether in prod or local dev. It would be nice if pyodide was always cached to speed up local dev, although I don't know if it actually makes a difference. For everything else, we don't want caching to interfere with normal non-service-worker development, so it should be off by default.

On that note, now that we're removing cross-origin isolation, the service worker is no longer optional and including it in development (without caching) needs to be more convenient. The instructions in https://github.com/alexmojaki/futurecoder#running-locally need to be updated to reflect that, and a command/script needs to be added to build and copy the service worker for the sake of `npm start`.

frontend/public/index.html
```diff
@@ -3,8 +3,7 @@
 <head>
     <meta charset="utf-8"/>
     <link rel="shortcut icon" type="image/png" href="/static/logo/favicon_32px.png"/>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-          integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
```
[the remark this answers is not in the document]
hangtwenty: I tried reverting to the previous way of including bootstrap then whitelisting that CDN in the `registerRoute`/`staleWhileRevalidate` configuration - [651d271](https://github.com/alexmojaki/futurecoder/pull/320/commits/651d271e58442d2f93a36b80b680949950ebf4ba). When going offline, Bootstrap doesn't come from cache, so it's a failing request. Might be something different about CORS in this case (CSS from Stackpath CDN, with `crossorigin="anonymous"`) versus the other cases (such as Pyodide from jsdelivr.net).

So for now I reverted the revert to keep the branch in a working state.

frontend/src/service-worker.js
hangtwenty: As for:

> On that note, now that we're removing cross-origin isolation, the service worker is no longer optional and including it in development (without caching) needs to be more convenient. The instructions in [alexmojaki/futurecoder#running-locally](https://github.com/alexmojaki/futurecoder#running-locally) need to be updated to reflect that, and a command/script needs to be added to build and copy the service worker for the sake of npm start.

Draft of this README change: [5809506](https://github.com/alexmojaki/futurecoder/pull/320/commits/5809506d03d7ab169e32f88eb82647bbbab82839). But I'm unclear on this so I reverted it too ([de2b4cf](https://github.com/alexmojaki/futurecoder/pull/320/commits/de2b4cfd255fa094c0c25779ca7807ad8d280b1a)).  We can re-apply the change if it's actually what you meant.

hangtwenty: I am going to leave the README update to you because I'm just guessing at what you have in mind. It seems like you have a clearer sense of what you want the README update to look like.

frontend/src/serviceWorkerRegistration.js
```diff
@@ -0,0 +1,136 @@
...
+    // [::1] is the IPv6 localhost address.
+    window.location.hostname === '[::1]' ||
+    // 127.0.0.0/8 are considered localhost for IPv4.
+    window.location.hostname.match(/^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/)
+);
+
+export function register(config) {
+  // XXX Unlike some usages of PWAs, we always register the service worker (including in dev);
+  // this is because our service worker has responsibilities besides caching. But as you can see
+  // in service-worker.js, we only do the caching parts when it's NODE_ENV=production (i.e. "build")
```
alexmojaki, line 14: comment needs updating

frontend/src/serviceWorkerRegistration.js
```diff
@@ -0,0 +1,136 @@
...
+                'New content is available and will be used when all ' +
+                  'tabs for this page are closed. See https://cra.link/PWA.'
+              );
+
+              // Execute callback
+              if (config && config.onUpdate) {
+                config.onUpdate(registration);
+              }
+            } else {
+              if (process.env.REACT_APP_PRECACHE === true) {
```
alexmojaki, line 75:
```suggestion
              if (process.env.REACT_APP_PRECACHE) {
```

Surely all the env vars are strings?

hangtwenty: Doh, of course. I used to use a library for .env files that would parsed booleans, but it was not dotenv and it was not even recent, so idk why I did this. Thanks for catching it.

alexmojaki:
Overall, it needs to be reasonably easy for someone to run the dev server locally, make changes, test them etc. as one would normally develop. There needs to be instructions somewhere, probably the README as it is now. Ideally it shouldn't be too hard to roughly guess what to do without instructions, it shouldn't be too easy to accidentally do the wrong thing causing confusion and frustration, and it should be somewhat convenient.

Right now the problems are:

1. `npm start` isn't enough on its own. It's already bad enough that people have to run `./generate.sh` to run code on the Python side. Now they also have to build the service worker and copy it into `public`. We need instructions that tell people that, and ideally a script/command to help them do it, something more convenient than `npm run build && cp course/service-worker.js public`. But since this operation is slow and infrequently needed, it shouldn't automatically run as part of something else more common. I glimpsed you adding `prereq` or something to package.json at some point, and that intrigued me, not sure if that's the right solution.
2. `npm run build` currently actually sets `REACT_APP_PRECACHE=true` which produces a service worker that isn't suitable for development. So the command above would need to be `craco build && cp ...`. It's too easy for someone to get this wrong and then wonder why their dev environment is acting weird. There's no easy way to remember the difference between `npm run build` and `craco build`.
3. There's these other cryptic commands `build:full` and `pwa`. What do they mean? When should they be used? What's with the ports 3333 and 4444? If they're worth keeping, there should be some kind of docs somewhere, e.g. advice for a workflow when changing the service worker. (on a related note, `test` and `eject` aren't really useful either)

How do you think we could make things clearer and easier? I don't think I've been doing a good job at making this repo developer friendly on my own, so I don't want to just slap on more changes.

alexmojaki:
OK, I've gone with my gut for the README.

I'm happy to merge this. Are you?

hangtwenty:
@alexmojaki the README and script updates make sense to me. I think it's ready for merge. Do you need me to test again or do you feel confident it's good to go?

alexmojaki:
Huge congratulations and thank you for all your hard work and patience!!

PR #325, opened 2022-03-31
WIP — configuring Workbox to precache birdseye 

hangtwenty:
### Context

This is a branch off of #320.

@alexmojaki [said:](https://github.com/alexmojaki/futurecoder/pull/320#issuecomment-1082872939)

> Finally, remember that birdseye is still completely broken in this PR, offline or not, and that has to be fixed. Precaching birdseye would also be nice but is less essential.

I [replied:](https://github.com/alexmojaki/futurecoder/pull/320#issuecomment-1084010824)

> Re: precaching birdseye, I _think_ we want to get Workbox to [pre]cache everything in `frontend/public`. Pretty sure that can be accomplished by configuring [Workbox's `InjectManifest` options, i.e. `include`](https://developer.chrome.com/docs/workbox/reference/workbox-build/#:~:text=One%20or%20more%20specifiers%20used%20to%20include%20assets%20in%20the%20precache%20manifest.). Currently those options are hidden inside craco's CRA config, and they can be overridden but I have not gotten it working yet.

----

### This branch

Try overriding craco's `WebpackWorkboxPlugin.InjectManifest` options.

Currently it results in this error:

```
> frontend@0.1.0 build <redacted>/futurecoder/frontend
> craco build

{
  swSrc: '<redacted>/futurecoder/frontend/src/service-worker.js',
  exists: true
}
Creating an optimized production build...
Failed to compile.

Can't find self.__WB_MANIFEST in your SW source.
```

... which is unexpected because `src/service-worker.js` definitely has `self.__WB_MANIFEST` in it, and it works fine when we don't override the config.


hangtwenty:
Closing this for now as we are focused on the other branch. Just wanted to share the diff and error message without breaking #320.

frontend/craco.config.js
```diff
@@ -1,10 +1,43 @@
...
   webpack: {
     configure: (webpackConfig, {env, paths}) => {
+      // Output to ./course (instead of ./build)
       paths.appBuild = webpackConfig.output.path = path.resolve('course');
+
+      // Override config for WorkboxWebpackPlugin.{GenerateSW,InjectManifest}
+      webpackConfig.plugins = webpackConfig.plugins.map(plugin => {
+        if(plugin.constructor.name === 'InjectManifest') {
+          console.info({plugin})
+          return new WorkboxWebpackPlugin.InjectManifest({
```
hangtwenty, line 17: So, doing the `WorkboxWebpackPlugin` override this way actually gets past that error I was seeing (the one about missing `self.__WB_MANIFEST`). So that's good.

frontend/craco.config.js
```diff
@@ -1,10 +1,43 @@
...
+            exclude: [/\.map$/, /asset-manifest\.json$/, /LICENSE/],
+            maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
+            /////////////////////////////////////////////////////////////////////////////////////////////////////////////
+            // To get birdseye to be included in precaching ... This might work ...
+            /////////////////////////////////////////////////////////////////////////////////////////////////////////////
+            include: [
+              // NOT SURE IF THIS IS ACTUALLY WORKING... Intention is to get all contents of "./public/",
+              // including birdseye stuff and bootstrap stuff, to be part of `precacheAndRoute(self.__WB_MANIFEST)`
+              // in service-worker.js. At time of writing, other options are Craco's defaults for WorkboxWebpackPlugin.
+              /public\/.*/
```
hangtwenty, line 32: However, I don't think this `include` argument is having an effect.

Now that we know how to override, it's possible to try tweaking this option and seeing if another way works.

hangtwenty:
I'm not sure if `include` will actually do the trick. Might be something more subtle with a (new/additional) `registerRoute` call...?

frontend/craco.config.js
```diff
@@ -1,12 +1,48 @@
...
 
 module.exports = {
-  // Output to ./course (instead of ./build)
   webpack: {
     configure: (webpackConfig, {env, paths}) => {
+      // Output to ./course (instead of ./build)
       paths.appBuild = webpackConfig.output.path = path.resolve('course');
       return webpackConfig;
     },
+    plugins: [
```
hangtwenty: Doing it at `plugins` rather than `webpack.configure` did not work.

<details><summary>Original error (no longer relevant)</summary>

[0be6cb3](https://github.com/alexmojaki/futurecoder/commit/0be6cb3970a07cf2e7830872a51bf32a477597d5) results in this error:

```
> frontend@0.1.0 build <redacted>/futurecoder/frontend
> craco build

{
  swSrc: '<redacted>/futurecoder/frontend/src/service-worker.js',
  exists: true
}
Creating an optimized production build...
Failed to compile.

Can't find self.__WB_MANIFEST in your SW source.
```

... which is unexpected because `src/service-worker.js` definitely has `self.__WB_MANIFEST` in it, and it works fine when we don't override the config.

</details>

hangtwenty:
btw @alexmojaki what do you think of the approach mentioned above (https://github.com/alexmojaki/futurecoder/pull/325#issuecomment-1084049537), i.e. "inlining" birdseye so that precaching it is more straightforward?

PR #341, opened 2022-04-27
Store user state with localForage for speed and offline usage

alexmojaki:
Closes https://github.com/alexmojaki/futurecoder/issues/163

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -15,6 +15,7 @@ import {
   moveStep,
   setDeveloperMode,
   setEditorContent,
+  signOut,
   specialHash,
 } from "./book/store";
 import Popup from "reactjs-popup";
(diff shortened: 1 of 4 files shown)


alexmojaki:
This allows futurecoder to be used while completely offline once it's been loaded for the first time. Caching of content/resources was done in #320 while this takes care of user data that was previously only persisted to firebase. It works even if you never connect to firebase, so the site now works in China and anywhere else where firebase is blocked. Closing and reopening or reloading should be fine, even while offline. Loading should also be faster since it doesn't have to wait for firebase to get started.

Things you can't do while offline:

- birdseye
- Python tutor
- Sign in if you weren't already

@hangtwenty can you do some extra testing on this, particularly while offline? That way your PWA work should also get some extra testing.

@saviosg @gaogaotiantian I expect you'll be interested in this too.

PR #362, opened 2022-06-26
Added chapter on sets

AndrewPluzhnikov:
Hello,

I wrote a chapter introducing sets.

Please let me know if you'd like me to change or add anything.

### DIFF
--- a/core/chapters/c06a_sets.py
+++ b/core/chapters/c06a_sets.py
@@ -1,0 +1,291 @@
+import ast
+import random
+from textwrap import dedent
+from typing import List
+
+from core.exercises import generate_list, generate_string
+from core.text import (
+    ExerciseStep,
+    MessageStep,
+    Page,
+    Step,
+    VerbatimStep,
+    search_ast,
+    Disallowed,
+)
+
+
+
+class IntroducingSets(Page):
+    class introduce_iterables(VerbatimStep):
+        """
+Before introducing sets, it's helpful to understand iterables.
+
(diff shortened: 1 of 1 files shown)


alexmojaki:
Hello,

Thank you for being willing to contribute to futurecoder! It's always great to have new contributors. I'm also very impressed that you learned how course content is implemented and wrote a whole chapter, that looks like a lot of work.

Unfortunately I don't think I can use this right now. I see several issues with how this is written. I don't want to list them all, but one example is that it uses many concepts that haven't been taught elsewhere in the course, such as:

- Multiple arguments to `range` and `print`
- Tuples
- Comments
- Imports
- Formatting

I mentioned in https://github.com/alexmojaki/futurecoder/blob/master/how_to_contribute.md#helping-with-course-content that I prefer for people to make a proposal draft in markdown before trying to implement anything. Maybe I should add some extra clarity/emphasis to that point. It's not just that implementing content is technically tricky. I also don't want people to go through the effort of implementing course content too soon, to prevent this kind of thing. I'm sorry that wasn't clear.

But there's a bright side! Since you've learned how to implement course content, you can help with #363. Adding a chapter on dicts is a much higher priority than sets. The content is already written, it just needs to be implemented. I was procrastinating doing that myself and didn't expect that anyone else would be able to, so having help would be great. Are you interested?

AndrewPluzhnikov:
Alex,
I'd like to go ahead and implement the chapter on dictionaries.
I will comment on the other issue(#363) with some questions.

Issue #363, opened 2022-06-27
Chapter about dicts


alexmojaki:
Part of https://github.com/alexmojaki/futurecoder/issues/165

I've written a partial draft of a chapter about dictionaries here: https://demo.hedgedoc.org/OH78A7uwTDOjeMM4v0QFLg?edit

It's not complete, but a lot has been written, and the actual implementation of the existing content can start now, although suggestions for changes are still welcome.

AndrewPluzhnikov:
Alex,
I've read the draft and I have a few suggestions.

- In the nested language dictionaries example, you use f-strings. I think we can avoid using these here - I don't think they are needed.
- Could we use quantity and quantities instead of amount and amounts? Amounts are generally used for uncountable values, while quantities can be used for both countable and uncountable values. 

I will go ahead and start implementing the draft.

alexmojaki:
This chapter is going to be added after all the other existing chapters, i.e. after the tic-tac-toe project. f-strings have been introduced and used several times, and I think it's best to keep using them.

Changing amount to quantity sounds fine.

Thanks for implementing!

alexmojaki:
Next work on this is in https://github.com/alexmojaki/futurecoder/pull/370#issuecomment-1311489751

PR #370, opened 2022-07-12
Added ch12 Dictionaries.

AndrewPluzhnikov:
Alex,

Here is approximately the first half of the Dictionaries chapter from issue #363 . 
Please let me know if I am making some mistakes or if you have any comments.

I also fixed a bug in exercises.py. When the input parameter is of type `Dict[str, str]`, running `generate.sh` resulted in the error below:
```
  File "C:\Program Files\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Program Files\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Andrew\futurecoder\translations\generate_po_file.py", line 163, in <module>
    main()
  File "C:\Users\Andrew\futurecoder\translations\generate_po_file.py", line 47, in main
    for step_name, text in zip(page.step_names, page.step_texts(raw=True)):
  File "C:\Users\Andrew\futurecoder\core\text.py", line 329, in step_texts
    result = [step.raw_text if raw else step.text for step in cls.steps[:-1]] + [cls.final_text]
  File "C:\Users\Andrew\futurecoder\core\text.py", line 372, in steps
    return [self.get_step(step_name) for step_name in self.step_names]
  File "C:\Users\Andrew\futurecoder\core\text.py", line 372, in <listcomp>
    return [self.get_step(step_name) for step_name in self.step_names]
  File "C:\Users\Andrew\futurecoder\core\text.py", line 325, in get_step
    clean_step_class(step)
  File "C:\Users\Andrew\futurecoder\core\text.py", line 143, in clean_step_class
    cls.test_exercise(cls.solution)
  File "C:\Users\Andrew\futurecoder\core\text.py", line 644, in test_exercise
    for inputs, result in cls.test_values():
  File "C:\Users\Andrew\futurecoder\core\text.py", line 637, in test_values
    inputs = cls.generate_inputs()
  File "C:\Users\Andrew\futurecoder\core\text.py", line 653, in generate_inputs
    return {
  File "C:\Users\Andrew\futurecoder\core\text.py", line 654, in <dictcomp>
    name: generate_for_type(typ)
  File "C:\Users\Andrew\futurecoder\core\exercises.py", line 130, in generate_for_type
    return {
KeyError: typing.Dict[str, str]
```


core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+            total_cost(
+                ['apple', 'box', 'cat'],
+                {'apple': 2, 'box': 5, 'cat': 100, 'dog': 100},
+                ),
+            107,
+        )
+        """ 
+
+        def solution(self):
+            def total_cost(cart: Dict[str, str], prices: Dict[str, str]):
```
alexmojaki, line 200:
```suggestion
            def total_cost(cart: List[str], prices: Dict[str, int]):
```

But in any case it doesn't work here to just rely on the type annotations. You need to write a custom `generate_inputs` so that the elements of `cart` are keys in `prices`.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+Note that curly brackets (`{}`) are used to create the dictionary instead of the square brackets (`[]`) used when writing lists.
+
+Remember that with lists, you get values based on their *index*, i.e. their position in the list.
+So if `words = ['apple', 'box']`, then `words[0]` is `'apple'` and `words[1]` is `'box'`.
+Try this in the shell:
+
+__program_indented__
+        """
+
+        auto_translate_program = False
```
alexmojaki, line 55: Any reason for this?

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+
+        program = "french[0]"
+
+        def check(self):
```
alexmojaki, line 59: No need for a custom `check` in a `VerbatimStep`. They've been told exactly what to write, so it will check that they wrote exactly that.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+
+    class dict_access3(VerbatimStep):
+        """
+That's better!
+
+Now run a similar line in the shell to look up the translation for 'box'.
+
+__program_indented__
+        """
+        ##TODO: Hide the program from the user but still let them proceed through the excercise
```
alexmojaki, line 87: If you remove `__program_indented__` and run `generate_po_file`, you should get an error saying `Either include __program__ or __program_indented__ in the text, or set program_in_text = False in the class.`. In this case, `program_in_text = False` is the way to solve this TODO.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+On the other hand, keys have to be unique. Imagine if your prices started like this:
+
+__program_indented__
+        """
+
+        auto_translate_program = False
+
+        program = "prices = {'apple': 2, 'apple': 3}"
+
+    class shopping_cart3(ExerciseStep):
```
alexmojaki, line 162: These first three steps are all meant to be one step. Students aren't meant to run just `prices = {'apple': 2, 'apple': 3}`, nothing would happen.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+
+And even though your customers have calculators in their pockets, they still expect you to add up all the prices
+yourself and tell them how much this will all cost, because that's what normal shops do.
+
+So let's write a function that does that. Complete the function below, particularly the line `price = ...`
+
+
+    __copyable__
+    __program_indented__
+
```
alexmojaki, line 181:
```suggestion
```

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+
+    class shopping_cart5(VerbatimStep):
+        """
+We've added another parameter called `quantities` to `total_cost`.
+Now `cart` is still a list of strings, but it doesn't have any duplicates.
+`quantities` is a dictionary where the keys are the items in `cart` and the corresponding values are the quantity
+of that item that the customer wants to buy.
+
+
```
alexmojaki, line 257: This line here is meant to be the boundary between steps, not the "We've added another..." line above. That text is part of the same step. In https://demo.hedgedoc.org/OH78A7uwTDOjeMM4v0QFLg?edit these boundaries are represented by `----`. Similarly the students are not required to run the original DNA program.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+                        char = 'G'
+                    result += char
+                return result
+
+            original = 'AGTAGCGTCCTTAGTTACAGGATGGCTTAT'
+            expected = 'TCATCGCAGGAATCAATGTCCTACCGAATA'
+            assert_equal(substitute(original), expected)
+
+
+    ##TODO: Is this supposed to be an exercise or a verbatim step?
```
alexmojaki, line 293: A VerbatimStep means "run this exact code", usually shown to the user. `shopping_cart4` and `dna_part2` need to be exercise steps because the user has to figure out what to write to fill in the `...`, and there are multiple possible correct answers.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+i.e. `reverse = swap_keys_values(letters)`."""
+
+class DictionaryKeysAndValues(Page):
+    class introducing_keys(Step):
+        """
+Copy this code into the editor, then change `print(quantities)` to `print(quantities.keys())`, and run the whole program.
+
+    __copyable__
+    __no_auto_translate__
+
```
alexmojaki, line 357:
```suggestion
```

Auto translation should work fine here. Also, I think it'd be better not to have blank lines between the special keywords like `__copyable__` and the code.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+    quantities = {'apple': 1, 'cat': 10}
+    print(quantities)
+
+
+        """
+
+        program = """
```
alexmojaki, line 364: In general I prefer `program` to be defined as a function rather than a string, unless the program is really short or has syntax errors.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+        quantities = {'apple': 1, 'cat': 10}
+        print(quantities.keys())
+        """
+
+        def check(self):
+            return "dict_keys(['apple', 'cat'])" in self.result
```
alexmojaki, line 370:
```suggestion
            return "dict_keys([" in self.result
```

This method of checking is a good idea here. This change is so that it will still work when the exercise is translated, whether that's automatic or not.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+
+        program = """
+        quantities = {'apple': 1, 'cat': 10}
+        for key in quantities:
+            print(key)
+                """
+
+        def check(self):
+            return "apple\ncat" in self.result
```
alexmojaki, line 401: Again, this won't work after translation. But also this check will pass even if the user just runs the original code without removing `.keys()`. I think just using a VerbatimStep is best here.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+        total_cost(
+            {'dog': 5000000, 'box': 2},
+            {'apple': 2, 'box': 5, 'cat': 100, 'dog': 100},
+        ),
+        500000010,
+    )
+        """
+
+        def solution(self):
+            def total_cost(quantities: Dict[str, str], prices: Dict[str, str]):
```
alexmojaki, line 428: Again, these type hints need correcting, but a `generate_inputs` is needed anyway.

core/chapters/c12_dictionaries.py
```diff
@@ -0,0 +1,510 @@
...
+    class english_to_french(ExerciseStep):
+        """
+That looks nice! We've fully solved the problem of adding up the total cost.
+
+Coming back to our first example: write a function
+which prints out each word in an English-to-French dictionary and its translation, labeling them with their languages.
+Here's your starting code:
+
+    __copyable__
+    __no_auto_translate__
```
alexmojaki, line 456: Even here, I think automatic translation will be fine. I expect the French translations for each word to be:

```
french -> anglais
apple -> pomme
pomme -> apple
box -> boite
boite -> box
```

alexmojaki:
Thank you! This is a good start. Almost all these comments are about things that are very easy to fix.

> I also fixed a bug in exercises.py. When the input parameter is of type Dict[str, str], running generate.sh resulted in the error below:

That's because that type of parameter has never appeared in exercises before! By the way, when you run `generate.sh`, the files it changes are meant to be committed and pushed.

core/exercises.py
```diff
@@ -121,10 +121,26 @@ def generate_list(typ):
...
         if typ.__origin__ is list:
             return generate_list(only(typ.__args__))
+        if typ.__origin__ is dict:
+            return {}
+=======
+        return {
+            list: generate_list,
+            dict: generate_dict,
+        }[typ.__origin__](*typ.__args__)
+>>>>>>> 0491f0843816b3cfe6e33da02d1fa45c984bd8b8
```
alexmojaki, line 143:
```suggestion
        return {
            list: generate_list,
            dict: generate_dict,
        }[typ.__origin__](*typ.__args__)
```

alexmojaki:
Looks like you're having git problems. I completely sympathise, I've had similar struggles and I still don't fully understand this stuff.   The easiest option might be to just create a new branch from master, copy in the files that you're actually changing (I think there's just two), and make a new PR.

core/chapters/c12_dictionaries.py
```diff
@@ -425,6 +432,10 @@ def total_cost(quantities: Dict[str, str], prices: Dict[str, str]):
 
         hints = "TODO PPPP"
 
+        # Custom generate inputs is required because cart and prices must have the same keys.
+        def generate_inputs():
+            return {'quantities': {"apple" : 2, "banana" : 5}, 'prices': {"apple" : 7, "banana" : 11}}
```
alexmojaki, line 437: `generate_inputs` should generate values that fit the required constraints but are otherwise random and will usually be different each time it's called. When the user's submission is checked, `generate_inputs` is called 10 times to create different test cases. This ensures that the user has really found a general solution and not just written a bunch of `if` statements that pass the initial tests. Static examples like this one belong in the `tests` list below.

alexmojaki:
> Sorry it took me so long.

No need to apologize, I appreciate the help.

Why the new PR in https://github.com/alexmojaki/futurecoder/pull/383? This one is identical and the remaining comments are still relevant.

alexmojaki:
It's my turn to apologise for slowness :sweat_smile: 

Let me know what you think of my changes. Otherwise there's a TODO that needs resolving, and I think the hints need to be fleshed out more.

alexmojaki:
Preview: https://futurecoder-ta--dicts-ai9rq6np.web.app/course/#toc

alexmojaki:
Thank you for your help!

Do you want to continue with the rest of the dicts chapter? Alternatively, I saw you started translating to Russian way back. I don't know why you chose to stop, but that would still be appreciated!

AndrewPluzhnikov:
Hi Alex,

I was wondering if you had any other draft chapters written up in markup
which I could try to work on?

Sincerely,
Andrew Pluzhnikov

alexmojaki:
Hi @AndrewPluzhnikov, I appreciate the enthusiasm to keep contributing. I've just finished another page on https://demo.hedgedoc.org/OH78A7uwTDOjeMM4v0QFLg?edit, here's a copy:

Now we'll learn how to add key-value pairs to a dictionary,
e.g. so that we can keep track of what the customer is buying.
Before looking at dictionaries, let's remind ourselves how to add items to a list. Run this program:

    __copyable__
    cart = []
    cart.append('dog')
    cart.append('box')
    print(cart)

------

Pretty simple. We can also change the value at an index, replacing it with a different one:

    __copyable__
    cart = ['dog', 'cat']
    cart[1] = 'box'
    print(cart)

// This step should have `predicted_output_choices`, with `['dog', 'box']` and `['box', 'cat']` as options.

------

What if we used that idea to create our list in the first place?
We know we want a list where `cart[0]` is `'dog'` and `cart[1]` is `'box'`, so let's just say that:

    __copyable__
    cart = []
    cart[0] = 'dog'
    cart[1] = 'box'
    print(cart)

------

Sorry, that's not allowed. For lists, subscript assignment only works for existing valid indices.
But that's not true for dictionaries! Try this: 

    quantities = {}
    quantities['dog'] = 5000000
    quantities['box'] = 2
    print(quantities)

Note that `{}` means an empty dictionary, i.e. a dictionary with no key-value pairs.
This is similar to `[]` meaning an empty list or `""` meaning an empty string.

// This step should have `predicted_output_choices`

-----

That's exactly what we need. When the customer says they want 5 million boxes,
we can just put that information directly into our dictionary. So as an exercise, let's make a generic version of that.
Write a function `buy_quantity(quantities)` which calls `input()` twice to get an item name and quantity from the user
and assigns them in the `quantities` dictionary. Here's some starting code:

    __copyable__
    def buy_quantity(quantities):
        print('Item:')
        item = input()
        print('Quantity:')
        ...

    def test():
        quantities = {}
        buy_quantity(quantities)
        print(quantities)
        buy_quantity(quantities)
        print(quantities)

    test()

and an example of how a session should look:

    Item:
    dog
    Quantity:
    5000000
    {'dog': 5000000}
    Item:
    box
    Quantity:
    2
    {'dog': 5000000, 'box': 2}

Note that `buy_quantity` should *modify* the dictionary that's passed in, and doesn't need to return anything.
You can assume that the user will enter a valid integer for the quantity.

// This also needs a MessageStep which is triggered if the user's solution is correct except that the values
// are still strings instead of ints.

-----

Well done!

Next exercise: earlier we defined a function `total_cost(quantities, prices)` which returned a single number
with a grand total of all the items in the cart. Now let's make a function `total_cost_per_item(quantities, prices)`
which returns a new dictionary with the total cost for each item:

    __copyable__
    def total_cost_per_item(quantities, prices):
        totals = {}
        for item in quantities:
            ... = quantities[item] * prices[item]
        return totals

    assert_equal(
        total_cost_per_item({'apple': 2}, {'apple': 3, 'box': 5}),
        {'apple': 6},
    )

    assert_equal(
        total_cost_per_item({'dog': 5000000, 'box': 2}, {'dog': 100, 'box': 5}),
        {'dog': 500000000, 'box': 10},
    )

-----

Perfect! This is like having a nice receipt full of useful information.

Let's come back to the example of using dictionaries for translation. Suppose we have one dictionary
for translating from English to French, and another for translating from French to German.
Let's use that to create a dictionary that translates from English to German:

    __copyable__
    def make_english_to_german(english_to_french, french_to_german):
        ...
        // The user has to write the whole function, here's an example solution:
        english_to_german = {}
        for english in english_to_french:
            french = english_to_french[english]
            german = french_to_german[french]
            english_to_german[english] = german
        return english_to_german

    assert_equal(
        make_english_to_german(
            {'apple': 'pomme', 'box': 'boite'},
            {'pomme': 'apfel', 'boite': 'kasten'},
        ),
        {'apple': 'apfel', 'box': 'kasten'},
    )

// this step should have `parsons_solution = True`

-----

Great job!

Of course, language isn't so simple, and there are many ways that using a dictionary like this could go wrong.
So...let's do something even worse! Write a function which takes a dictionary and swaps the keys and values,
so `a: b` becomes `b: a`.

    __copyable__
    def swap_keys_values(d):
        ...

    assert_equal(
        swap({'apple': 'pomme', 'box': 'boite'}),
        {'pomme': 'apple', 'boite': 'box'},
    )

-----

Magnificent!

Jokes aside, it's important to remember how exactly this can go wrong. Just like multiple items in the store
can have the same price, multiple words in English can have the same translation in French.

But there are many situations where you can be sure that the values in a dictionary *are* unique and that this
'inversion' makes sense. For example, we saw this code [earlier in the chapter](#UsingDictionaries):
    
    __copyable__
    __no_auto_translate__
    def substitute(string, d):
        result = ""
        for letter in string:
            result += d[letter]
        return result
    
    plaintext = 'helloworld'
    encrypted = 'qpeefifmez'
    letters = {'h': 'q', 'e': 'p', 'l': 'e', 'o': 'f', 'w': 'i', 'r': 'm', 'd': 'z'}
    reverse = {'q': 'h', 'p': 'e', 'e': 'l', 'f': 'o', 'i': 'w', 'm': 'r', 'z': 'd'}
    assert_equal(substitute(plaintext, letters), encrypted)
    assert_equal(substitute(encrypted, reverse), plaintext)

Now we can construct the `reverse` dictionary automatically:

    reverse = swap_keys_values(letters)

For this to work, we just have to make sure that all the values in `letters` are unique.
Otherwise it would be impossible to decrypt messages properly. If both `'h'` and `'j'` got replaced with `'q'`
during encryption, there would be no way to know whether `'qpeef'` means `'hello'` or `'jello'`!

Issue #395, opened 2022-10-23
Relax the dependecy over python3.10.2


vincentxavier:
`pyproject.toml` imposes a hard dependency on a specific Python version. Maybe we could lower it.

alexmojaki:
Unfortunately not, it's to match the Python version used in Pyodide, which is what the Python code actually runs in. Packaging dependencies and running tests shouldn't happen in a different version.

vincentxavier:
Just have a look at https://forge.aeif.fr/futurecoder/futurecoder.forge.aeif.fr/-/compare/main...pyproject and https://fr.futurecoder.io/course/#IntroducingDictionaries

PR #396, opened 2022-10-23
Relax the Python version constrain

vincentxavier:
See issue #395


alexmojaki:
You can see the tests failing due to differences between 3.10.2 and 3.10.8. It's annoying, but that's why I specified 3.10.2 in https://github.com/alexmojaki/futurecoder/commit/92a4c3f66335696a8318ff808132c0aee52582a6. Other python versions will probably work, probably even 3.9, but they are not officially tested and recommended.

PR #419, opened 2023-01-07
Fix exception cause in exercises.py

cool-RR:
More context for this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

### DIFF
--- a/core/exercises.py
+++ b/core/exercises.py
@@ -27,11 +27,11 @@ def make_function(program, arg_names):
             target = only(node.targets)
             assert isinstance(target, ast.Name)
             assert target.id == arg_name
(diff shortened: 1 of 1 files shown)


alexmojaki:
Thanks, I remember seeing that blog post, I think it's good to educate developers about this and encourage the standard.

However, this particular exception is being used as control flow - it's caught immediately here: https://github.com/alexmojaki/futurecoder/blob/be97985f128b98c39622a129f358173381c976d2/core/text.py#L643

So I think `from e` might make the code less clear by implying that the exception bubbles up and shows a traceback that should be readable.

PR #465, opened 2024-02-18
Ruff fixes

cclauss:
https://docs.astral.sh/ruff is the modern linter and formatter for Python code.

This pull request applies a few [lint rules](https://docs.astral.sh/ruff/rules/) to help teach good practices to new coders.

% `ruff --statistics`
```
42	F405	[ ] `AnimationGroup` may be undefined, or defined from star imports
 6	F401	[*] `core.text.search_ast` imported but unused
 5	F541	[*] f-string without any placeholders
 2	E402	[ ] Module level import not at top of file
 2	F841	[*] Local variable `french` is assigned to but never used
 1	E701	[ ] Multiple statements on one line (colon)
 1	F403	[ ] `from manim import *` used; unable to detect undefined names
```
% `ruff rule F401`
# unused-import (F401)

Derived from the **Pyflakes** linter.

Fix is sometimes available.

## What it does
Checks for unused imports.

## Why is this bad?
Unused imports add a performance overhead at runtime, and risk creating
import cycles. They also increase the cognitive load of reading the code.

If an import statement is used to check for the availability or existence
of a module, consider using `importlib.util.find_spec` instead.

If an import statement is used to re-export a symbol as part of a module's
public interface, consider using a "redundant" import alias, which
instructs Ruff (and other tools) to respect the re-export, and avoid
marking it as unused, as in:

```python
from module import member as member
```

## Example
```python
import numpy as np  # unused import


def area(radius):
    return 3.14 * radius**2
```

Use instead:
```python
def area(radius):
    return 3.14 * radius**2
```

To check the availability of a module, use `importlib.util.find_spec`:
```python
from importlib.util import find_spec

if find_spec("numpy") is not None:
    print("numpy is installed")
else:
    print("numpy is not installed")
```

## Options
- `lint.ignore-init-module-imports`

## References
- [Python documentation: `import`](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement)
- [Python documentation: `importlib.util.find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec)
- [Typing documentation: interface conventions](https://typing.readthedocs.io/en/latest/source/libraries.html#library-interface-public-and-private-symbols)

### DIFF
--- a/core/chapters/c04_for_loops.py
+++ b/core/chapters/c04_for_loops.py
@@ -16,7 +16,8 @@ __program_indented__
 
         def program(self):
             name = 'World'
-            for character in name: print(character)
+            for character in name:
+                print(character)
 
     final_text = """
 You can read the code almost like normal English:
--- a/core/chapters/c04_for_loops.py
+++ b/core/chapters/c04_for_loops.py
@@ -16,8 +16,7 @@ __program_indented__
 
         def program(self):
             name = 'World'
-            for character in name:
-                print(character)
+            for character in name: print(character)
 
     final_text = """
 You can read the code almost like normal English:
--- a/core/chapters/c01_the_shell.py
+++ b/core/chapters/c01_the_shell.py
@@ -1,6 +1,5 @@
 # flake8: NOQA E501
 import ast
-from textwrap import dedent
 
 from core.text import MessageStep, Page, Step, VerbatimStep, search_ast
 
--- a/core/chapters/c05_if_statements.py
+++ b/core/chapters/c05_if_statements.py
@@ -5,7 +5,7 @@ from textwrap import dedent
 
 import pure_eval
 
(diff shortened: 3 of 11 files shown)


core/chapters/c04_for_loops.py
```diff
@@ -16,7 +16,8 @@ class first_for_loop(VerbatimStep):
 
         def program(self):
             name = 'World'
-            for character in name: print(character)
```
alexmojaki, line 19: This is intentional as the user hasn't learned about indented blocks yet.

core/chapters/c04_for_loops.py
```diff
@@ -16,7 +16,8 @@ class first_for_loop(VerbatimStep):
...
+            for character in name:
+                print(character)
```
cclauss, line 20:
```suggestion
            for character in name: print(character)
```

